PR #42, opened 2015-01-27
Ability to pass file object to TextBuffer and update pathwatcher for mor...

mostafaeweda:
* Introduce a new parameter file to TextBuffer to allow injecting custom implementation of the file (e.g. a remote `File`)
* Remove the normal flow dependency on the `getDigest` being a synchronous function and introduce the `getDigestSync` for the (de)serialization usage

Dependency: https://github.com/atom/node-pathwatcher/pull/56 - that's the package.json reference

### DIFF
--- a/spec/text-buffer-spec.coffee
+++ b/spec/text-buffer-spec.coffee
@@ -1,5 +1,6 @@
 {existsSync, moveSync, readFileSync, removeSync, writeFileSync} = require 'fs-plus'
 {join} = require 'path'
+Q = require 'q'
 temp = require 'temp'
 {File} = require 'pathwatcher'
 TextBuffer = require '../src/text-buffer'
@@ -990,14 +991,16 @@ describe "TextBuffer", ->
 
       it "notifies ::onDidConflict observers", ->
         buffer.setText("a change")
-        buffer.save()
-        buffer.insert([0, 0], "a second change")
-
         handler = jasmine.createSpy('fileChange')
-        writeFileSync(filePath, "a disk change")
-        buffer.onDidConflict handler
+        buffer.save().then ->
+          buffer.insert([0, 0], "a second change")
+          buffer.onDidConflict handler
+          expect(handler.callCount).toBe 0
+          # wait a little bit to not confuse the save change event, with the write file change event
+          setTimeout ( ->
+            writeFileSync(filePath, "a disk change")
+          ), 500
 
-        expect(handler.callCount).toBe 0
         waitsFor ->
           handler.callCount > 0
 
@@ -1119,13 +1122,12 @@ describe "TextBuffer", ->
         expect(buffer.isModified()).toBe true
 
         modifiedHandler.reset()
-        buffer.save()
+        buffer.save().then ->
+          expect(modifiedHandler).toHaveBeenCalledWith(false)
+          expect(buffer.isModified()).toBe false
+          modifiedHandler.reset()
 
-        expect(modifiedHandler).toHaveBeenCalledWith(false)
-        expect(buffer.isModified()).toBe false
-        modifiedHandler.reset()
-
-        buffer.insert([0, 0], 'x')
+          buffer.insert([0, 0], 'x')
 
       waitsFor ->
         modifiedHandler.callCount is 1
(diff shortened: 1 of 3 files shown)


mostafaeweda:
@kevinsawicki currently,
So, would you please bump the `node-pathwatcher` to npm (that includes https://github.com/atom/node-pathwatcher/pull/56), so I can include its new version here in the `package.json` instead of the github tarball master commit reference

kevinsawicki:
> Would you please bump the node-pathwatcher to npm

Just published 4.0.0

mostafaeweda:
@kevinsawicki I just fixed the specs here with text-buffer, but let's chat anytime about the scenario of this workaround: https://github.com/mostafaeweda/text-buffer/commit/6ee56b252da10c0e869893257bfdc99c619f306a#diff-d2de3ac7c7201920cebd82c128081211R1000

src/text-buffer.coffee
```diff
@@ -1214,10 +1236,9 @@ class TextBuffer
   #                {File::read} (default: false).
   # * `callback`   (optional) {Function} to call after the cached contents have
```
kevinsawicki: :fire: since this param is gone

mostafaeweda:
@kevinsawicki this is the last PR on our short-term critical path forward - other stuff will come incrementally and probably non-pushing

mostafaeweda:
@kevinsawicki any update on this?

kevinsawicki:
Having only an async save option on TextBuffer needs to be investigated more.

I'm worried that when Atom is quit, and it prompts to save unsaved editors, an async save might not complete before the render process is terminated.

Have you checked what happens in this case?

mostafaeweda:
@kevinsawicki I just tested this now a couple of times and it seems working just fine.

spec/text-buffer-spec.coffee
```diff
@@ -990,14 +991,16 @@ describe "TextBuffer", ->
...
-        writeFileSync(filePath, "a disk change")
-        buffer.onDidConflict handler
+        buffer.save().then ->
+          buffer.insert([0, 0], "a second change")
+          buffer.onDidConflict handler
+          expect(handler.callCount).toBe 0
+          # wait a little bit to not confuse the save change event, with the write file change event
+          setTimeout ( ->
+            writeFileSync(filePath, "a disk change")
+          ), 500
```
kevinsawicki: Could elaborate a bit more on why this timeout was needed?

src/text-buffer.coffee
```diff
@@ -320,6 +323,22 @@ class TextBuffer
   getPath: ->
     @file?.getPath()
 
+  # Public: Set the associated file for the buffer.
+  #
+  # * `file` A {File} representing the file
+  setFile: (file) ->
+    return if file?.getPath() == @getPath()
```
kevinsawicki: Why shouldn't you be able to set the file when the buffer's path matches the file's?

mostafaeweda: right, I should be able to

spec/text-buffer-spec.coffee
mostafaeweda: every save emits a watcher change and if two saves happens really fast:
* The first is a good normal save, triggered by the IDE
* The second is conflicting non-editor-related conflict that happened almost the same time the edited the text buffer, then it's a conflict - but the change watcher was emitted twice almost the same time and checked and emitted two conflict events for one real conflict.

I was actually waiting on your help here to get this right, because this didn't happen when the save was synchronous.

kevinsawicki: Okay, I'll have to take a look at this test a bit deeper and think it through

mostafaeweda: Actually, with more investigation, I found out the the synchronous save in node emits one change event, but the async one emits two change events!!!! - pending more investigation

mostafaeweda: @kevinsawicki I just tested it, `fs.writeFile` causes two `change` events while `fs.writeFileSync` causes only one `change` event

bolinfest:
I think that we need to get https://github.com/atom/atom/issues/5706 fixed first.

Though from a cursory look through those PRs, I think it might be possible to upgrade safely without breaking those packages because they seem to pull in their own `pathwatcher` dependencies and instantiate `File` from those rather than operate on a `File` that was created by Atom core. Assuming that is the case, then Atom core could use a different version of `File` than those packages.

alexandernst:
Any news on this one? It does sound quite interesting to be able to remotely edit files.

nathansobo:
@bolinfest is this one still something you want to proceed with? Kinda forgot about it.

PR #43, opened 2015-01-29
Adds "install": "grunt" script and moves grunt to dependencies from devDependencies

secrettriangle:
Necessary because package.json lists "main" as a script which doesn't exist until text-buffer is built!

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "description": "A container for large mutable strings with annotated regions",
   "main": "./lib/text-buffer",
   "scripts": {
+    "install": "grunt",
     "prepublish": "grunt clean lint coffee shell:update-atomdoc atomdoc",
     "test": "grunt test"
   },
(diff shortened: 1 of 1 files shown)


kevinsawicki:
Could you explain a bit more about why this is needed?

The prepublish task compiles all the .coffee to .js so `npm install text-buffer` shouldn't require grunt at all.

What issue are you seeing?

secrettriangle:
Well, I'm using [a fork of text-buffer](https://github.com/slap-editor/text-buffer) in my project, so when I use a github URL (like `"text-buffer": "slap-editor/text-buffer"`), the `lib/` directory is not present.

You can see how I'm using text-buffer here: `npm install -g secrettriangle/slap#text-buffer`

kevinsawicki:
Yeah, npm doesn't run `prepublish` scripts when you use a repository dependency so the files aren't compiled. See https://github.com/npm/npm/issues/3055

I don't think shipping grunt to work-around this issue is something we should do since it greatly bloats the package, basically this affects all libraries that use CoffeeScript and compile on `prepublish`.

You can work around it in your fork by checking in the compiled `lib/` folder that `grunt` generates.

PR #44, opened 2015-01-29
Adds Point#negate and Range#negate

secrettriangle:


kevinsawicki:
Would you mind adding some tests for this new method?

secrettriangle:
Sure! Should I make a new `spec/point-spec.coffee` file, or include it in an existing file?

kevinsawicki:
If you think `Range.negate` would be helpful as well, then I'd say add it and just test it for both points and ranges through `range-spec.coffee`.

But if you don't think this method will also be useful for `Range`s then a new spec file (`point-spec.coffee`) is good.

secrettriangle:
I've updated the PR:

* Added Range#negate
* Created point-spec.coffee
* Testing Range#negate in range-spec.coffee

kevinsawicki:
Thanks for adding the specs!

kevinsawicki:
Publishes in 4.1.0

PR #46, opened 2015-02-12
add did-delete event:

kans:
Adds did-delete event and dispatcher as well as a simple unit test.

Currently, plugins must rely on the file's did-delete event, but the underlying file can change or itself be deleted which makes the existing implementation cumbersome.

### DIFF
--- a/spec/text-buffer-spec.coffee
+++ b/spec/text-buffer-spec.coffee
@@ -1049,6 +1049,21 @@ describe "TextBuffer", ->
         expect(bufferToDelete.getPath()).toBe filePath
         expect(bufferToDelete.isModified()).toBeFalsy()
 
+
+    describe "when the file is deleted", ->
+      events = []
+      beforeEach ->
+        bufferToDelete.onDidDelete -> events.push true
+        deleteHandler = jasmine.createSpy('deleteHandler')
+        bufferToDelete.onDidDelete deleteHandler
+        removeSync(filePath)
+        waitsFor "file to be deleted", ->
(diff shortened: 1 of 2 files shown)


spec/text-buffer-spec.coffee
```diff
@@ -1011,6 +1011,7 @@ describe "TextBuffer", ->
       filePath = join(temp.dir, 'atom-file-to-delete.txt')
       writeFileSync(filePath, 'delete me')
       bufferToDelete = new TextBuffer({filePath, load: true})
+      bufferToDelete.onDidDelete -> didDelete = true
```
kevinsawicki: Is this `didDelete` variable use anywhere?

spec/text-buffer-spec.coffee
```diff
@@ -1049,6 +1050,22 @@ describe "TextBuffer", ->
...
         expect(bufferToDelete.isModified()).toBeFalsy()
 
+
+    describe "when the file is deleted", ->
+      events = []
+      beforeEach ->
+        bufferToDelete.onDidDelete -> events.push true
+        deleteHandler = jasmine.createSpy('deleteHandler')
+        bufferToDelete.onDidDelete deleteHandler
+        removeSync(filePath)
```
kevinsawicki: Just curious, why is `removeSync` called here and also down below in the `it block?

kans:
@kevinsawicki 

Good catch. I removed the offending code and rebased the branch.

kevinsawicki:
Thanks for this, sorry for the delay merging it.

PR #47, opened 2015-02-13
Exists sync

bolinfest:


bolinfest:
@kevinsawicki I believe you can bump this to 4.1.5 after this. Then we should bump `text-watcher@4.1.5` in Atom. That should eliminate all of the deprecated calls to `File.exists()` in Atom core.

kevinsawicki:
Thanks :shipit:

PR #51, opened 2015-03-11
:arrow_up: pathwatcher ^4.1.2

bolinfest:


nathansobo:
@bolinfest Looks like this fails. Wish I would have noticed before upgrading a bunch of other packages to 4.x because now we have 2 versions on master. :grimacing:

nathansobo:
I'm reverting all the pull requests and going back to 3.x. Please reopen them when we have a green build on this.

bolinfest:
I put together https://github.com/atom/node-pathwatcher/pull/64 to add a `writeSync()` method on the 4.x branch.

I guess the safer thing would be to add it on the 3.x branch, as well, and then it should hopefully be a straight dependency bump here?

kevinsawicki:
Travis is having some issues, but the specs pass for me locally so I'm going to merge this, thanks @bolinfest!

PR #52, opened 2015-03-17
Grouped undo not updating marker ranges

kevinsawicki:
Looks like grouped undo isn't properly updating marker ranges as shown in the test case attached to this PR.

This came about from reducing https://github.com/atom/bracket-matcher/issues/110 which I believe will be fixed by fixing the test case here.

The basic problem is that marker ranges aren't getting updated on undo operations causing the marker range to be past the end of the buffer which is causing various issues in Atom since it allows cursors/selections to exceed the buffer's range.

/cc @maxbrunsfeld @nathansobo


nathansobo:
This didn't have anything to do with time based grouping, but it was a problem with restoring marker state after undoing transactions. Our existing approach made no sense. We've updated the behavior to always record the state of all markers at the start of a transaction so it can be restored afterward.

It's a pretty big change that could have some subtle implications for packages... here's what will be slightly different:

Basically, when handling an undo event for a text change, the state of markers will reflect that change. Then there could be a subsequent marker update event that restores them to their original state when the transaction finishes being undone. Everything is still technically correct and as long as authors are observing marker changes there should be no problem. However, some marginal use cases could encounter issues. We'll run the bundled package specs before upgrading to make sure we're at least clear on our side. It's worth watching out for over the coming days

/cc @izuzak (We can do a hangout to explain this better if that paragraph wasn't clear. It's kinda hard to explain this)

kevinsawicki:
> This didn't have anything to do with time based grouping

Oh interesting, I could only reproduce it when using transactions (both in Atom, and in specs)

nathansobo:
@kevinsawicki We're just going to push this to master and fight the :fire: if there is one. It's too hard to do anything else with semver.

kevinsawicki:
Could you hold off on publishing until I get https://github.com/atom/atom/pull/6011 merged?

Sorry, just want to power through any pathwatcher weirdness with that upgrade.

nathansobo:
Ended up undoing this merge... it was premature. Will open another PR and reference this one soon.

PR #54, opened 2015-03-26
Add an `onWillChange` event to `TextBuffer`

benogle:


src/text-buffer.coffee
```diff
@@ -628,6 +645,10 @@ class TextBuffer
     endRow = oldRange.end.row
     rowCount = endRow - startRow + 1
 
+    changeEvent = {oldRange, newRange, oldText, newText}
```
benogle: We cool doing this once for both events?

kevinsawicki: You mean using the same object for both events? I'd say each should get their own event object in case the listener modifies it.

benogle: Yeah, I was on the fence. two objects it is

PR #60, opened 2015-05-12
Document Point and Range instance properties

mnquintana:
Part of https://github.com/atom/atom/issues/6714


mnquintana:
The failing specs were failing on master locally for me, so I don't think this changed anything there.

PR #64, opened 2015-05-25
Supply `getRange` to `scanInRange` iterator

as-cii:
While scanning a large file, we may not want to compute the position for every result. This is especially important for searches where there’s an elevated amount of matches and we’re interested only in a subset of them.

This commit introduces a `getRange` parameter in the `iterator` function, which is suited to be a drop-in replacement for `range`. The logic is the same, but passing a function, or a getter if you will, enables us to perform all the position math (i.e. `positionForCharacterIndex` is rather expensive) lazily. :racehorse:

This is still “slow”, however, because we cannot delete the `range` parameter without first changing all the scan methods' clients to use `getRange` instead. Thus, we need to work incrementally by shipping both versions (eagerly and lazily computed `range`) and remove the eagerly computed one as soon as we complete the migration.

After that, we will benefit from a significant performance boost thereby solving bracket-matcher slowness when matching parentheses in a reasonably long file.

### Core Repos To Be Updated

* [ ] [atom-autocomplete](https://github.com/atom/autocomplete/tree/master/lib/autocomplete-view.coffee) (we may probably skip it, as we use autocomplete-plus now)
* [ ] [find-and-replace](https://github.com/atom/find-and-replace)
* [ ] [bracket-matcher](https://github.com/atom/bracket-matcher)
* [ ] [atom](https://github.com/atom/atom)

/cc: @nathansobo @maxbrunsfeld @atom/feedback

### DIFF
--- a/src/text-buffer.coffee
+++ b/src/text-buffer.coffee
@@ -976,13 +976,18 @@ class TextBuffer
       matchStartIndex = match.index
       matchEndIndex = matchStartIndex + matchLength
 
-      startPosition = @positionForCharacterIndex(matchStartIndex + lengthDelta)
-      endPosition = @positionForCharacterIndex(matchEndIndex + lengthDelta)
-      range = new Range(startPosition, endPosition)
+      getRange = =>
+        startPosition = @positionForCharacterIndex(matchStartIndex + lengthDelta)
+        endPosition = @positionForCharacterIndex(matchEndIndex + lengthDelta)
+
+        new Range(startPosition, endPosition)
+
+      # FIXME: Remove this as soon as all clients use `getRange` in their iterators.
+      range = getRange()
       keepLooping = true
       replacementText = null
       matchText = match[0]
-      iterator({ match, matchText, range, stop, replace })
+      iterator({ match, matchText, range, stop, replace, getRange })
 
       if replacementText?
         @setTextInRange(range, replacementText)
--- a/src/text-buffer.coffee
+++ b/src/text-buffer.coffee
@@ -40,13 +40,15 @@ class SearchResultCallback
     @stopped is false
 
   getRange: =>
+    return @computedRange if @computedRange?
+
     matchStartIndex = @match.index
     matchEndIndex = matchStartIndex + @matchText.length
 
     startPosition = @buffer.positionForCharacterIndex(matchStartIndex + @lengthDelta)
     endPosition = @buffer.positionForCharacterIndex(matchEndIndex + @lengthDelta)
 
-    new Range(startPosition, endPosition)
+    @computedRange = new Range(startPosition, endPosition)
 
 class TransactionAbortedError extends Error
   constructor: -> super
--- a/src/text-buffer.coffee
+++ b/src/text-buffer.coffee
@@ -39,7 +39,7 @@ class SearchResultCallback
   keepLooping: ->
     @stopped is false
 
-  getRange: ->
+  getRange: =>
     matchStartIndex = @match.index
     matchEndIndex = matchStartIndex + @matchText.length
 
--- a/src/text-buffer.coffee
+++ b/src/text-buffer.coffee
@@ -13,7 +13,7 @@ MarkerStore = require './marker-store'
 Patch = require './patch'
 {spliceArray, newlineRegex} = require './helpers'
 
-class SearchResultCallback
+class SearchResultIteratee
   Object.defineProperty @::, "range",
     get: ->
       return @computedRange if @computedRange?
@@ -1003,11 +1003,11 @@ class TextBuffer
 
     matches.reverse() if reverse
     for match in matches
-      callback = new SearchResultCallback(this, match, lengthDelta)
-      iterator(callback)
-      lengthDelta += callback.getReplacementDelta() unless reverse
+      resultIteratee = new SearchResultIteratee(this, match, lengthDelta)
+      iterator(resultIteratee)
+      lengthDelta += resultIteratee.getReplacementDelta() unless reverse
 
-      break unless global and callback.keepLooping()
+      break unless global and resultIteratee.keepLooping()
     return
 
   # Public: Scan regular expression matches in a given range in reverse order,
--- a/src/text-buffer.coffee
+++ b/src/text-buffer.coffee
@@ -13,23 +13,13 @@ MarkerStore = require './marker-store'
 Patch = require './patch'
 {spliceArray, newlineRegex} = require './helpers'
 
-class SearchMatch
-  Object.defineProperty @prototype, "range", get: ->
-    matchStartIndex = @match.index
-    matchEndIndex = matchStartIndex + @getMatchLength()
-
-    startPosition = @buffer.positionForCharacterIndex(matchStartIndex + @lengthDelta)
-    endPosition = @buffer.positionForCharacterIndex(matchEndIndex + @lengthDelta)
-
-    new Range(startPosition, endPosition)
+class SearchResultCallback
+  Object.defineProperty @::, "range", get: -> @getRange()
 
   constructor: (@buffer, @match, @lengthDelta) ->
     @stopped = false
     @replacementText = null
     @matchText = @match[0]
-
-  getMatchLength: ->
-    @matchText.length
 
   hasReplacedText: ->
     @replacementText?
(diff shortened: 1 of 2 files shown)


as-cii:
The benchmarks provided in the screenshots below show the potential of the proposed changes.

*(Tested on 55k LOC file, with a parenthesis at the beginning and the matching one at the end. Think to `atom/benchmark/fixtures/huge.js` but with the content between the parentheses 5 times larger)*

### Before

before

### After

after

:racehorse: :racehorse: :racehorse: 

When the file is larger the performance impact is even more evident :chart_with_upwards_trend:

lee-dohm:
Third-party packages may be using `scanInRange` as well. If we are going to hold to the API freeze, will we be able to remove the `range` parameter?

as-cii:
> Third-party packages may be using scanInRange as well. If we are going to hold to the API freeze, will we be able to remove the range parameter?

That's a fair question: thank you for bringing this up @lee-dohm :bow:

If that's the constraint, we might as well pass a parameter to tell this function to explicitly skip the eager position computation; namely, I am thinking to change `scanInRange`'s signature to something like:

```coffee
scanInRange: (regex, range, iterator, reverse=false, computePositionsEagerly = false) ->
```

I am :cry: because this feels a bit awkward to say the least, given that the boolean doesn't *really* change this function's behavior. On the other hand we won't break clients and they can always choose to opt-in this improvement gradually (to the point that we'll be able to remove the boolean).

Thoughts? :thought_balloon:

lee-dohm:
I'm :cry: too because this is a huge win as far as performance goes. The only alternative I've been able to think of is to create a new `scanInRangeLazy` function (stupid name, I know), but that just clutters up the API. I think your solution is the best one that doesn't break the API.

maxbrunsfeld:
I'd suggest moving all our packages to use `getRange`, but defining `range` as a lazily-computed property via `Object.defineProperty`, so that third property packages can still use it. Maybe at some we could deprecate it.

I'm not sure what the cost is of calling `Object.defineProperty` in a tight loop like this. If it's significant, we could introduce a class for these event objects, and only define the property once: on the prototype.

nathansobo:
We should absolutely move to a property getter on the prototype of the event object. Should be completely drop in unless I'm missing something.

as-cii:
Thanks for your suggestions, @nathansobo and @maxbrunsfeld. Eventually, I went with an event object which defines `range` as a *computed property* on the prototype. Everything is still snappy :racehorse: 

However, as suggested by @maxbrunsfeld, I also included `getRange` as a public method: I have mixed feelings about this, but ultimately I think it's clearer if clients just use `getRange` instead, as to remark that we may have some computation in there and to avoid [assignments to that variable](https://github.com/atom/find-and-replace/blob/master/lib/select-next.coffee#L92) (which from now on will be simply ignored and things may not work as expected).

I think the roadmap for this should be:

* :shipit: and bump version of `text-buffer`
* :arrow_up: `text-buffer` on `atom/atom`, migrating clients as needed
* Migrate all the other clients that I have listed in this PR's body, bumping their version
* :arrow_up: such clients on `atom/atom`
* Release Atom :rocket: 

What do you think? :thought_balloon:

lee-dohm:
Sounds like a good plan to me! Thanks for so elegantly addressing my concerns about the API :grinning:

src/text-buffer.coffee
```diff
@@ -13,6 +13,40 @@ MarkerStore = require './marker-store'
 Patch = require './patch'
 {spliceArray, newlineRegex} = require './helpers'
 
+class SearchResultCallback
```
nathansobo: Not sure about the name "callback" as it implies a function to me, whereas this is an object. What about just `SearchResult`? Or are you looking for some other word to indicate the control methods it provides?

maxbrunsfeld:
> avoid assignments to that variable

Nice catch.

> Migrate all the other clients that I have listed in this PR's body, bumping their version.

Usually, we wait to use features in packages until they are in a released version of Atom. That way, the packages' CI builds can continue to pass against the latest Atom release, and if any issues come up that delay our release of Atom, we can continue to release new package versions.

So I would say that we should release this change in a completely backward-compatible manner before changing any package code. To deal with the case that you caught above, where `find-and-replace` assigned to the `range` property, we should add a `set` descriptor to the `range` property that sets the `computedRange` memoized value. I think that keep `find-and-replace` working as it did before.

I still think we should add `getRange` function for the reasons you mentioned above, but we should release a new version of Atom before we migrate any packages to use that function.

src/text-buffer.coffee
```diff
@@ -965,30 +999,13 @@ class TextBuffer
...
-      range = new Range(startPosition, endPosition)
-      keepLooping = true
-      replacementText = null
-      matchText = match[0]
-      iterator({ match, matchText, range, stop, replace })
-
-      if replacementText?
-        @setTextInRange(range, replacementText)
-        lengthDelta += replacementText.length - matchLength unless reverse
+      callback = new SearchResultCallback(this, match, lengthDelta)
```
nathansobo: Same with this local variable. Can we call this something else like `result`?

nathansobo:
As for adding the `getRange` method, I see it this way:

Pros:
* Consistency: Other things on that object are methods
* Clarity: It's clearer that accessing the range could involve computation

Cons:
* Duplication: We have to leave the `range` property for backward compatibility, so now we have 2 ways to access it on this object.
* API churn: If we want to eliminate the duplicated property someday we would need to push people to use the method before 2.0 someday in the future.

I can get behind moving to `getRange` because it's nicer and more consistent, but it's not without downsides.

maxbrunsfeld:
> I can get behind moving to getRange because it's nicer and more consistent, but it's not without downsides.

I agree. I'd be ok with either adding it or not adding it.

src/text-buffer.coffee
as-cii: > Not sure about the name "callback" as it implies a function to me, whereas this is an object.

Agreed. :+1: 

> Or are you looking for some other word to indicate the control methods it provides?

Exactly. I'd expect a `SearchResult` to be a simple value object, whereas this one provides a bunch of features which seem to be outside the scope of an object which merely holds data.

Also, I would stray a bit from `SearchResult` as it may imply that we return the result of the whole search, whilst what we're really passing along is a *search occurrence*.

I am thinking to something along the lines of `SearchOccurrence`, but that clearly expresses that it provides control over the iteration as well. `SearchOccurrenceIteratee` maybe? :confused:

as-cii:
API churn is definitely a major concern for me and I agree with @nathansobo that having two ways of doing the same thing is a bit awkward. On the other hand, I dislike having a field whose behavior is a bit obscure for all the reasons I stated above.

Since we still need to keep backwards compatibility, I'll go on and simply remove `getRange` and introduce a setter, as suggested.

as-cii:
I went ahead and renamed `SearchResult` to `SearchResultIteratee` as well as adding a setter for `range`. I have tested it against `find-and-replace` and everything seems to work smoothly.

I'd :heart: to have your feedback on this, especially about this last renaming operation. After that I think we can :shipit:! :tada:

nathansobo:
What about `SearchCallbackArgument`. It's super literal to the point of being a bit lame, but also very explicit.

as-cii:
> What about SearchCallbackArgument. It's super literal to the point of being a bit lame, but also very explicit.

It's ok but I feel quite good about the concept of iteratee and I believe it fits quite well this use case. I see a callback as a fire and forget kind of a thing, whereas an iteratee implies the knowledge of what an iteration is, providing ways to control it (e.g. stopping or replacing, in this case).

However, I can change it to `SearchCallbackArgument` if you think it suits this object better :+1:

nathansobo:
Is "iteratee" a word you've seen used elsewhere? It sounds a bit odd to me.

as-cii:
> Is "iteratee" a word you've seen used elsewhere? It sounds a bit odd to me.

I think it comes from the functional programming world, but I have seen it used in [underscorejs](http://underscorejs.org) as well. On the other hand, [MDN shows no result for `iteratee`](https://developer.mozilla.org/en-US/search?q=iteratee).

At the end of the day, we'll have to maintain this and I am ok if we want to go for a more literal solution as the one you have proposed rather than introducing some programming jargon which we may not understand in the future. :+1: 

I have renamed it in 662553725d907f486d829a8e9c011bb19b6dd3d8: feel free to jump on this for further refinements :art: 

Thank you everyone for the feedback provided so far! :bow:

nathansobo:
@as-cii Okay I just researched "iteratee" a bit, and it's not immediately clear to me it's the same concept as this argument we're passing to the iterator. But I see your reason for naming it that as the thing being iterated over. Still, I think especially since this is a fairly formal concept from FP, perhaps we should avoid the name unless we are totally sure we have the same semantics. I agree it would be nice to name this concept though.

nathansobo:
:heart: Thanks for investigating and fixing this issue!

maxbrunsfeld:
Great job as usual, @as-cii! This'll ship in Atom 0.202.

as-cii:
:tada: :tada: :tada: 

Thank you for your support guys :bow:

PR #69, opened 2015-06-02
Marker Index Low Hanging Fruits

as-cii:
These straightforward, yet effective changes improve performance by ~2s when inserting text on 1000 different cursors (~ 25% improvement). We cannot claim we're fast yet, but we're getting there :raised_hands: 

| Before | After  |
| --------- | -------|
| screen shot 2015-06-02 at 21 25 39 | screen shot 2015-06-02 at 21 25 33

One more improvement I could think of right now would be to defer [marker invalidation](https://github.com/atom/text-buffer/blob/master/src/marker-store.coffee#L100-L119) and perform it at the end of a transaction: this should give another significant boost to `MarkerStore`, which after that should just take ~250ms instead of 1s (i.e. when typing a letter for 1000 cursors). This appears to be quite disruptive, though, and may require some additional thinking, therefore I decided to leave it for a later moment.

An additional pair of :eyes: is most welcome, therefore please do not hesitate to leave your feedback. Thanks! :bow: 

/cc: @maxbrunsfeld @nathansobo

### DIFF
--- a/src/marker-index.coffee
+++ b/src/marker-index.coffee
@@ -381,12 +381,16 @@ class MarkerIndex
       Range(start, @getEnd(id))
 
   getStart: (id) ->
-    if entry = @rangeCache[id]
-      entry.start ?= @rootNode.getStart(id)
+    return unless @rootNode.ids.has(id)
+
+    entry = @rangeCache[id] ?= templateRange()
+    entry.start ?= @rootNode.getStart(id)
 
   getEnd: (id) ->
-    if entry = @rangeCache[id]
-      entry.end ?= @rootNode.getEnd(id)
+    return unless @rootNode.ids.has(id)
+
+    entry = @rangeCache[id] ?= templateRange()
+    entry.end ?= @rootNode.getEnd(id)
 
   findContaining: (start, end) ->
     containing = new Set
@@ -439,6 +443,7 @@ class MarkerIndex
 
   dump: ->
     unless @rangeCacheFresh
+      @rootNode.ids.forEach (id) => @rangeCache[id] ?= templateRange()
       @rootNode.dump(Point.ZERO, @rangeCache)
       @rangeCacheFresh = true
     @rangeCache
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
:+1: Nice!

benogle:
You're a machine

src/set-helpers.coffee
```diff
@@ -6,7 +6,10 @@ setEqual = (a, b) ->
   true
 
 subtractSet = (set, valuesToRemove) ->
-  valuesToRemove.forEach (value) -> set.delete(value)
+  if set.size > valuesToRemove.size
```
nathansobo: Very clever.

src/marker-index.coffee
```diff
@@ -462,6 +466,4 @@ assertValidId = (id) ->
     throw new TypeError("Marker ID must be a string")
 
 templateRange = ->
-  range = new Range(Point.ZERO, Point.ZERO)
-  range.start = range.end = null
-  range
+  Object.create(Range.prototype)
```
nathansobo: Dig.

src/marker-index.coffee
```diff
@@ -381,12 +381,16 @@ class MarkerIndex
       Range(start, @getEnd(id))
 
   getStart: (id) ->
-    if entry = @rangeCache[id]
-      entry.start ?= @rootNode.getStart(id)
+    return unless @rootNode.ids.has(id)
```
nathansobo: This one is surprising. Any commentary?

maxbrunsfeld: I think that whereas before, we could use the id's presence in the `rangeCache` to know that the marker existed, we now create the `rangeCache` entries lazily, so we have to check the `rootNode` to know if the marker exists.

nathansobo: Of course.

nathansobo: Thought it was an optimization in and of itself.

PR #70, opened 2015-06-05
Preserve correct marker snapshots in undo/redo stack

maxbrunsfeld:
Now, whenever we perform a transaction, we push two `Checkpoints` to the undo history: one before and one after.

/cc @nathansobo

### DIFF
--- a/src/history.coffee
+++ b/src/history.coffee
@@ -1,3 +1,5 @@
+_ = require 'underscore-plus'
+
 SerializationVersion = 2
 
 class Checkpoint
@@ -24,10 +26,13 @@ class History
   groupChangesSinceCheckpoint: (checkpointId) ->
     checkpointIndex = @getCheckpointIndex(checkpointId)
     return false unless checkpointIndex?
+    hasSeenChanges = false
     for entry, i in @undoStack by -1
       break if i is checkpointIndex
       if @undoStack[i] instanceof Checkpoint
-        @undoStack.splice(i, 1)
+        @undoStack.splice(i, 1) if hasSeenChanges
+      else
+        hasSeenChanges = true
     true
 
   applyCheckpointGroupingInterval: (checkpointId, groupingInterval) ->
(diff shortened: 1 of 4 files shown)


src/history.coffee
```diff
@@ -24,60 +26,73 @@ class History
   groupChangesSinceCheckpoint: (checkpointId) ->
     checkpointIndex = @getCheckpointIndex(checkpointId)
     return false unless checkpointIndex?
+    hasSeenChanges = false
     for entry, i in @undoStack by -1
       break if i is checkpointIndex
       if @undoStack[i] instanceof Checkpoint
-        @undoStack.splice(i, 1)
+        @undoStack.splice(i, 1) if hasSeenChanges
```
nathansobo: So this allows us to preserve checkpoints that follow changes. Is the preserved checkpoint always at the top of the stack? If so, would it be possible to just start the loop at the penultimate stack entry to preserve the checkpoint at the top?

src/history.coffee
```diff
@@ -118,13 +133,18 @@ class History
...
-    snapshot = fromStack[checkpointIndex].snapshot
-    for entry in fromStack.splice(checkpointIndex) by -1
+    snapshotBefore = null
+    snapshotAfter = fromStack[checkpointIndex].snapshot
+    splicedEntries = fromStack.splice(checkpointIndex)
+    for entry in splicedEntries by -1
       toStack?.push(entry)
-      changes.push(entry) unless entry instanceof Checkpoint
-    {changes, snapshot}
+      if entry instanceof Checkpoint
```
nathansobo: Again, could this be simplified by assuming the `snapshotBefore` is at the top of the stack?

src/text-buffer.coffee
```diff
@@ -843,19 +842,15 @@ class TextBuffer
 
   # Public: Undo the last operation. If a transaction is in progress, aborts it.
   undo: ->
-    markerSnapshot = _.extend(@markerStore.createSnapshot(), @lastMarkerSnapshot)
-    if pop = @history.popUndoStack(markerSnapshot)
+    if pop = @history.popUndoStack(@markerStore.createSnapshot())
```
nathansobo: Maybe not now, but I wonder if you could save any time by creating the snapshot with knowledge of the existing undo/redo snapshot you're going to assign into via `_.defaults`. You would only need to do work for markers that weren't already present.

src/text-buffer.coffee
```diff
@@ -875,22 +870,24 @@ class TextBuffer
...
       @transactCallDepth--
 
-    @history.groupChangesSinceCheckpoint(checkpoint)
-    @history.applyCheckpointGroupingInterval(checkpoint, groupingInterval)
+    @history.groupChangesSinceCheckpoint(checkpointBefore)
+    checkpointAfter = @history.createCheckpoint(@markerStore.createSnapshot(false, true))
 
-    @lastMarkerSnapshot = @markerStore.createSnapshot(false, true)
+    now = Date.now()
+    @history.setCheckpointGroupingInterval(checkpointAfter, now, groupingInterval)
```
nathansobo: Could `set` and `apply` be combined?

src/history.coffee
maxbrunsfeld: There could be an arbitrary number of checkpoints at the top of the stack, all of which we probably want to preserve. So conceptually, what we're looking for are checkpoints that occur between changes. It's true that we could probably always make this loop one iteration shorter by starting at the penultimate entry, but I think we'd still need to keep some piece of state like `hasSeenChanges`.

src/history.coffee
maxbrunsfeld: In the case of `redo`, we want `snapshotBefore` to be the snapshot taken *right* before the change, such that it will then become `snapshotAfter` if we `undo` again.

So if you had manually created some checkpoints via `::createCheckpoint()`, the redo stack might look like this:

```
... [checkpoint] [CHANGE] [CHANGE] [checkpoint] [checkpoint] [checkpoint] [checkpoint]
         ^- snapshotAfter                ^- snapshotBefore
```

maxbrunsfeld: After writing this, I'm thinking maybe their names should be reversed. I was thinking of 'before' and 'after' in terms of before and after the in-progress operation is applied, but in terms of their position in the stack, it's the opposite.

src/history.coffee
nathansobo: Right. Because of nested transactions.

maxbrunsfeld:
Refs https://github.com/atom/atom/issues/6915

PR #72, opened 2015-06-08
Optimize buffer search

maxbrunsfeld:
* [x] Get existing tests passing against the optimized implementation.
* [x] Add random tests for `::scanInRange()` and `::backwardsScanInRange()`, since their implementation is a little more complex now.
* [x] Make sure that packages that uses these methods (e.g. `bracket-matcher`) work in all cases.

#### Problem

The `scanInRange()` method allows the caller to stop the scan by calling the `stop()` function from within the scan callback. Currently though, we scan the entire buffer range and collect the matches in an Array before calling the callback. This causes scan operations to be slow for large files, even if the caller is only looking for the first match. 

Here's a flame graph for moving the cursor to the end of the word in a 2.3MB file. There's a bunch of garbage collection pauses due to buffering the matches. The operation took **250ms**.

screen shot 2015-06-08 at 2 31 43 pm

#### Solution

We call the scan callback as soon as we find each match, so that if `stop()` is called, we can avoid any further calls to `RegExp.prototype.exec()`.

It's a little tricky for `backwardsScanInRange()`. Whereas before we simply scanned the entire range and then reversed the array of matches, we now need to break the range into chunks and scan each chunk. If no matches are found within a chunk, we need to expand the chunk backwards, in case matches straddle chunk boundaries.

Here's a flame graph for the same operation as above:

screen shot 2015-06-08 at 2 32 29 pm

It took **12ms**. You can now hold down `alt-right` in a big file without locking the UI.

### DIFF
--- a/spec/text-buffer-spec.coffee
+++ b/spec/text-buffer-spec.coffee
@@ -2,6 +2,7 @@
 {join} = require 'path'
 temp = require 'temp'
 {File} = require 'pathwatcher'
+Random = require 'random-seed'
 TextBuffer = require '../src/text-buffer'
 SampleText = readFileSync(join(__dirname, 'fixtures', 'sample.js'), 'utf8')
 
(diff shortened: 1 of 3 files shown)


maxbrunsfeld:
This also improves `bracket-matcher`'s performance when moving the cursor around a large HTML file. For a 6300 line file (the source of this page), `BracketMatcherView::updateMatch` used to take about 2 ms per call. Now it's down to about 0.7 ms.

PR #80, opened 2015-06-30
Allow matches at beginning of scanned range in TextBuffer.prototype.backwardsScanInRange

nathansobo:
In atom/atom#5806, there's a [hack](https://github.com/atom/atom/pull/5806/files#diff-24f2fe2bd60cb4194c3cf7d107fa0f99R414) when moving to the previous subword boundary to account for the fact that it's impossible to match a regex against the start of the buffer when scanning backwards in a range. This seems unfortunate and unnecessary, especially since we allow regex matches that end at the end of the range. On the other hand, this could cause subtle issues by slightly widening the matches admitted by this method.

@maxbrunsfeld: What do you think?

### DIFF
--- a/spec/text-buffer-spec.coffee
+++ b/spec/text-buffer-spec.coffee
@@ -1906,6 +1906,32 @@ describe "TextBuffer", ->
         expect(matches[2][1]).toBe 'rr'
         expect(ranges[2]).toEqual [[5,6], [5,13]]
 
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
I agree it should be symmetrical with the behavior of matches at the end of the range. This seems good to me.

PR #82, opened 2015-07-21
Add note about creating Points with one argument

mnquintana:
Just discovered that you can create a Point from one argument or a 1-element Array, which is currently undocumented.

/cc @atom/feedback

### DIFF
--- a/src/point.coffee
+++ b/src/point.coffee
@@ -10,6 +10,15 @@
 # new Point(1, 2)
 # [1, 2] # Point compatible Array
 # ```
+#
+# You can also create a Point by specifying just a row number, either by 
+# instantiating a new Point with one argument or with a Point-compatible 
+# 1-element array. Points created this way will have a `column` value of `0`.
+#
+# ```coffee
+# new Point(1) # Creates a Point at row 1, column 0
+# [1] # Creates a Point at row 1, column 0
+# ```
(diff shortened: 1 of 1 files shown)


benogle:
cc @nathansobo

nathansobo:
This is an accidental feature that some have used in some places. Question is whether it makes sense to make it official. It's probably too late to change the behavior.

mnquintana:
FWIW, we've used it in bundled code too (I'll edit when I find the lines)

It is a rather convenient shorthand, but I don't know if we want to make it official.

nathansobo:
@maxbrunsfeld what do you think here. We just need to decide on something.

* I'd rather be explicit everywhere and always specify the column and fix our packages, but how many packages already copy our antipatterns.
* On the other hand, it's not the end of the world to be slightly less explicit. It's work to change this now that it's out there, so we might as well document it.

maxbrunsfeld:
Yeah, I think that I tried changing this at one point and realized that I couldn't because of existing uses. I'd actually prefer to *not* document this behavior though, and fix any places that are relying on it in core or bundled packages.

nathansobo:
@mnquintana Thanks for bringing this up. If anyone is interested in correcting existing usages to make things more explicit, that would be excellent.

PR #87, opened 2015-07-28
Add option to save a backup file before saving buffer

nathansobo:
Refs atom/atom#3158

When writing directly to an existing file, it's possible to truncate the original file if an I/O error occurs during the write. This PR adds the option to save a backup file to the path of the original file with a `~` appended. If any exceptions occur, we restore from the backup and delete it.

I'm honestly not super well-versed in all the nuances of safely writing files... I'm interested in input on this. I realize that buffering the entire file through memory before writing the backup isn't that awesome, but I'm not sure there's a synchronous cross-platform alternative. I'm really wishing `saveAs` was asynchronous right now...

Also: Does anyone know if Node's `writeFileSync` method does an `fsync` automatically?

### DIFF
--- a/spec/text-buffer-spec.coffee
+++ b/spec/text-buffer-spec.coffee
@@ -1759,13 +1759,16 @@ describe "TextBuffer", ->
 
         describe "if a file already exists at the default backup path", ->
           it "chooses a different backup file name to avoid overwriting the existing file", ->
+            alternateBackupFilePath = backupFilePath + '~'
+            fs.removeSync(alternateBackupFilePath) if fs.existsSync(alternateBackupFilePath)
+
             fs.writeFileSync(backupFilePath, 'Contents of file at default backup path')
 
             saveAsBuffer.saveAs(filePath, backup: true)
             expect(fs.readFileSync(filePath, 'utf8')).toBe 'Buffer contents'
             expect(fs.readFileSync(backupFilePath, 'utf8')).toBe 'Contents of file at default backup path'
             expect(fs.existsSync(backupFilePath)).toBe true
-            expect(fs.existsSync(backupFilePath + '~')).toBe false
+            expect(fs.existsSync(alternateBackupFilePath)).toBe false
 
 
       describe "if no file exists at the path", ->
--- a/spec/text-buffer-spec.coffee
+++ b/spec/text-buffer-spec.coffee
@@ -1676,7 +1676,7 @@ describe "TextBuffer", ->
         expect(buffer.isInConflict()).toBeFalsy()
         expect(buffer.getText()).toBe(fileContents)
 
-  describe "::saveAs(path)", ->
+  describe "::saveAs(path, {backup})", ->
     [filePath, saveAsBuffer] = []
 
     afterEach ->
@@ -1723,6 +1723,46 @@ describe "TextBuffer", ->
 
       waitsFor ->
         changeHandler.callCount > 0
+
+    describe "if the 'backup' option is true", ->
+      [backupFilePath, saveAsBuffer] = []
+
+      beforeEach ->
+        filePath = join(temp.dir, 'temp.txt')
+        saveAsBuffer = new TextBuffer()
+        saveAsBuffer.setText 'Buffer contents'
+        backupFilePath = filePath + '~'
+
+      describe "if there is an existing file at the specified path", ->
+        beforeEach ->
+          fs.writeFileSync(filePath, 'File contents')
+
+        describe "if the file is written cleanly", ->
+          it "deletes the backup copy after writing", ->
+            saveAsBuffer.saveAs(filePath, backup: true)
+            expect(fs.readFileSync(filePath, 'utf8')).toBe 'Buffer contents'
+            expect(fs.existsSync(backupFilePath)).toBe false
+
+        describe "if an exception is thrown while writing to the file", ->
+          it "restores the file's original contents from the backup copy, deletes the backup copy, and re-throws the exception", ->
+            saveAsBuffer.setPath(filePath)
+            originalWriteSync = saveAsBuffer.file.writeSync
+            saveAsBuffer.file.writeSync = (args...) ->
+              originalWriteSync.apply(saveAsBuffer.file, args)
+              throw new Error('Something broke')
+
+            expect(-> saveAsBuffer.saveAs(filePath, backup: true)).toThrow 'Something broke'
+
+            expect(fs.readFileSync(filePath, 'utf8')).toBe 'File contents'
+            expect(fs.existsSync(backupFilePath)).toBe false
+
+      describe "if no file exists at the path", ->
+        it "does not make a backup before writing", ->
+          fs.removeSync(filePath) if fs.existsSync(filePath)
+          expect(fs.existsSync(filePath)).toBe false
+          expect(fs.existsSync(backupFilePath)).toBe false
+          saveAsBuffer.saveAs(filePath, backup: true)
+          expect(fs.existsSync(backupFilePath)).toBe false
 
   describe "::getTextInRange(range)", ->
     beforeEach ->
--- a/spec/text-buffer-spec.coffee
+++ b/spec/text-buffer-spec.coffee
@@ -1,10 +1,10 @@
-{existsSync, moveSync, readFileSync, removeSync, writeFileSync} = require 'fs-plus'
+fs = require 'fs-plus'
 {join} = require 'path'
 temp = require 'temp'
 {File} = require 'pathwatcher'
 Random = require 'random-seed'
 TextBuffer = require '../src/text-buffer'
-SampleText = readFileSync(join(__dirname, 'fixtures', 'sample.js'), 'utf8')
+SampleText = fs.readFileSync(join(__dirname, 'fixtures', 'sample.js'), 'utf8')
 
 describe "TextBuffer", ->
   buffer = null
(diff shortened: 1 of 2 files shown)


alexandernst:
Here you go another thing you might want to work on. INstead of saving the entire file, try writing only the bits of the file that actually changed. This will tackle https://github.com/atom/atom/issues/8091 automatically.

maxbrunsfeld:
> try writing only the bits of the file that actually changed

This would only be useful in very specific cases, right? Because AFAIK even if a single character is inserted at the beginning of the file, the entire file needs to be overwritten.

alexandernst:
@maxbrunsfeld Indeed, that would help only in some cases. Also, I'm not really sure how Atom handles writing currently. Are you using some custom methods or do you go low level?

spec/text-buffer-spec.coffee
```diff
@@ -1713,17 +1713,69 @@ describe "TextBuffer", ->
...
+          it "chooses a different backup file name to avoid overwriting the existing file", ->
+            fs.writeFileSync(backupFilePath, 'Contents of file at default backup path')
+
+            saveAsBuffer.saveAs(filePath, backup: true)
+            expect(fs.readFileSync(filePath, 'utf8')).toBe 'Buffer contents'
+            expect(fs.readFileSync(backupFilePath, 'utf8')).toBe 'Contents of file at default backup path'
+            expect(fs.existsSync(backupFilePath)).toBe true
+            expect(fs.existsSync(backupFilePath + '~')).toBe false
+
+
```
maxbrunsfeld: :fire: extra line break

maxbrunsfeld:
> Does anyone know if Node's writeFileSync method does an fsync automatically?

The `fs` module does expose an [`fsyncSync`](https://nodejs.org/api/fs.html#fs_fs_fsyncsync_fd) method, but you need to pass the file descriptor, which means we'd need to use `fs.writeSync` instead of `fs.writeFileSync`.

Looking at the native code for `fs`, the only mentions of the `fysnc(2)` function are in the [implementations of `fs.fynsc` and `fs.fsyncSync`](https://github.com/joyent/node/blob/d13d7f74d794340ac5e126cfb4ce507fe0f803d5/src/node_file.cc#L666), so I don't think that `writeFileSync` and friends do this automatically.

nathansobo:
> so I don't think that writeFileSync and friends do this automatically.

Seems like our [documentation](https://github.com/atom/fs-plus/blob/master/src/fs-plus.coffee#L257) on fs-plus is probably wrong if my interpretation of the word "flush" is correct. Seems like I should probably use the raw file handle API.

src/text-buffer.coffee
```diff
@@ -1199,19 +1200,33 @@ class TextBuffer
...
 
     @emitter.emit 'will-save', {path: filePath}
     @emit 'will-be-saved', this if Grim.includeDeprecatedAPIs
     @setPath(filePath)
-    @file.writeSync(@getText())
+
+    if options?.backup and @file.existsSync()
+      backupFilePath = filePath + '~'
+      backupFilePath += '~' while fs.existsSync(backupFilePath)
+      fs.writeFileSync(backupFilePath, @file.readSync())
```
bronson: It's worth using [fdatasync](http://linux.die.net/man/2/fdatasync) to make sure the backup file is on disk before truncating the main file.  Otherwise there's probably a couple-second window where, if you lose power, the user will find that the backup file isn't on disk yet and the main file's metadata is truncated.  Total loss.

Doing this after calling writeFileSync should be good enough:

```js
fd = fs.openSync(path, 'w')
fs.fdatasyncSync(fd) 
fs.closeSync(fd)
```

If you just created the backup file, you'll need to ensure its dirent is on disk too.  Otherwise, there's no point to ensuring the file contents are written!

```js
path = require 'path'
parent = path.dirname(@file.getPath())
fd = fs.openSync(parent, 'r')  // 'r' is important, 'w' doesn't work for dirs
fs.fdatasyncSync(fd) 
fs.closeSync(fd)
```

Just stumbled across this, matches my findings: http://www.daveeddy.com/2013/03/26/synchronous-file-io-in-nodejs/

src/text-buffer.coffee
```diff
@@ -1199,19 +1200,33 @@ class TextBuffer
...
+
+    try
+      @file.writeSync(@getText())
+      fs.removeSync(backupFilePath) if backupFilePath?
+    catch error
+      if backupFilePath?
+        fs.writeFileSync(filePath, fs.readFileSync(backupFilePath))
```
bronson: And, of course, before removing the backup file, you need to ensure the main file is fsync'd to disk.  Same technique as above: fsync the data, and fsync the directory if the file was just created.

bronson:
I left some pseudocode to fsync the files in an inline comment. Tell me if you want me to expand that into a clean PR. Should be pretty straightforward.

Syncing the data is required on posix and windows, syncing the directory is only needed on posix. (I assume fsyncing a dir is a no-op on windows but I've never tried.)

src/text-buffer.coffee
bronson: Why `fs.readFileSync(backupFilePath)`?  The text is still in `@getText()` isn't it?

src/text-buffer.coffee
```diff
@@ -1199,19 +1200,33 @@ class TextBuffer
...
-    @file.writeSync(@getText())
+
+    if options?.backup and @file.existsSync()
+      backupFilePath = filePath + '~'
+      backupFilePath += '~' while fs.existsSync(backupFilePath)
+      fs.writeFileSync(backupFilePath, @file.readSync())
+
+    try
+      @file.writeSync(@getText())
+      fs.removeSync(backupFilePath) if backupFilePath?
```
bronson: And, of course, before removing the backup file, you need to ensure the main file is fsync'd to disk.  Same technique as above: fsync the data, and fsync the directory if the file was just created.

src/text-buffer.coffee
```diff
@@ -1199,19 +1200,33 @@ class TextBuffer
...
+    catch error
+      if backupFilePath?
+        fs.writeFileSync(filePath, fs.readFileSync(backupFilePath))
+        fs.removeSync(backupFilePath)
```
bronson: Is the idea that you'll only get here if the file was successfully written the second time?  This seems risky, especially since we're not quite sure why the first attempt failed.  Why not just leave the backup file on disk?

src/text-buffer.coffee
nathansobo: Here I'm trying to restore the former contents of the file, as if nothing had ever been written to it. The value of `@getText()` is the new contents of the file that failed to write. Am I missing something? Do you think I shouldn't try to restore the contents of the file from the backup? It seems nice if we can avoid forcing people to manually fall back to the backup file.

src/text-buffer.coffee
nathansobo: Yeah, the idea is that we'd get here if we successfully restore the contents of the file from the backup. Obviously I'd need to do the same synchronization here as well... I was thinking that if our attempt to restore from the backup fails, we'll never reach this line and so the backup will get left around, but if we can restore cleanly then we'll clean it up. What holes are you seeing here?

nathansobo:
@bronson Thanks for the helpful comments... I really appreciate it!

src/text-buffer.coffee
nathansobo: Ah, this is kind of a bummer because `File.prototype.writeSync` has some conveniences related to privilege escalation... Can I just open a file descriptor for this path and `fdatasync` on it while allowing the write to proceed down the existing path, or do the write and the sync need to occur on the same file descriptor. I'm thinking they probably do, but thought I'd check before I go through the hassle.

bronson: Opening the file again (even if you open it 'r', though I'd certainly open it 'w') and calling fsync on that is accepted practice.  Even if you get a different fd, it still ensures that the blocks that were previously dirtied are written.  It smells a little funky but it should work fine.

src/text-buffer.coffee
bronson: Right, that makes sense.  I just assumed you were doing something different and read it my way.  :)

I figured you were writing the *new* contents to the backup file...  That way, if the main write fails, you still have the new contents in the backup file.  And, if the backup write fails, you still have the old contents in the main file.  When users hit save, they don't care about the old contents anymore, right?

Either way sounds fine to me.

src/text-buffer.coffee
bronson: In theory I think you're absolutely right.  In practice, once filesystem error paths start getting taken, I've seen some really weird things happen.  As long as it's an extremely rare occurrence, my tendency is to have the code back quickly away and leave as many pieces intact as it can.  The user will be happy to clean up.

nathansobo:
@bronson Okay, pretty sure this addresses everything you pointed out. Thanks again. It would have taken us much more time (and probably pain) to get this right without your expertise. :bow:

bronson:
Looks great!  Just two more thoughts:

It looks like you're going to have two filehandles open for writing on the same file at the same time.

```coffeescript
        fd = fs.openSync(filePath, 'w') # We open fd so we can call fdatasync
        @file.writeSync(@getText()) # File::writeSync does privelege escalation, so we don't use fd here
```

This will work fine on all OSs/filesystems I've used in the last decade but I remember computers being cranky about this.  Maybe writeSync before openSync?

And, you sure you want to keep adding ~ to find a free backup file?  Personally I like the idea of blowing away the existing backup file but, if that worries you, then I'd suggest at least putting a limit on the number of ~s you add.

nathansobo:
> Maybe writeSync before openSync?

When I tried to `openSync` with a `w` flag *after* the `writeSync`, it truncated the file. It worked fine with `r`, but since you suggested using `w` I decided to move it earlier. Maybe opening it `r` after the `writeSync` call is the lesser of two evils?

> And, you sure you want to keep adding ~ to find a free backup file?

I'm just worried about blowing away a file someone cares about that happens to end in `~` for some other random reason. Maybe I'm being paranoid. The additional `~`s will be finite, though I guess we could end up with a growing list of dead backup files if someone repeatedly attempts to write such that backups succeed but writing to the main file doesn't. My preference is to keep it this way because I'm worried about clobbering something someone wants.

bronson:
Dammit, I forgot that 'w' would truncate the file.

> Maybe opening it r after the writeSync call is the lesser of two evils?

Probably.  'a' might be good too.

Vim gets around the stale swapfile issue by stopping the world and making you deal with it before even opening the file.  Emacs, of course, [lets you do whatever you want](http://www.gnu.org/software/emacs/manual/html_node/elisp/Backup-Files.html) but, by default, I think it overwrites backup files once when it starts, then leaves them alone.

Your current behavior sounds fine but I'd really suggest a limit, even if it's in the tens of thousands... otherwise a bad interaction with a plugin or something, especially if it's on a *fat or older filesystem, could leave your files in a directory that can't be accessed anymore.

nathansobo:
> Your current behavior sounds fine but I'd really suggest a limit

Maybe I could throw an exception after 10 or something and saying that we couldn't save because no unused path exists for a backup file? Seems like the only option. We could potentially catch this exception in the notifications package and provide a UI to remediate the situation.

bronson:
That would be great.  With the limit in place, as far as I'm concerned, this can't be merged fast enough.

nathansobo:
Getting some weird failures on Travis... `illegal access` and the specs just terminate. I'm not pulling up much information on the Google about them.

maxbrunsfeld:
Those failures existed before intermittently.

PR #94, opened 2015-08-28
Don't save to existing location if not modified

jamesgpearce:
Holding down cmd-S too long will cause multiple synchronous saves which can be expensive if there is heavyweight linter activity listening for the save events.

This simply debounces it (in a sense) by not saving it the file if it is known not to have changed. Also this has no effect on saving new files for the first time.


jamesgpearce:
bounced travis and all is good... please review!

50Wliu:
Please add specs to this :)

PR #97, opened 2015-09-04
Always use serialized text when deserializing

maxbrunsfeld:
I realized that the fix added in #95 doesn't always work. When the file was saved, the undo history was still lost on reload. This fixes that.

* [ ] Since undo stack is now preserved when reopening a window, it is more likely to get long enough to cause memory usage issues. Introduce some fixed maximum length for the undo stack.

Depends on https://github.com/atom/node-pathwatcher/pull/85

### DIFF
--- a/src/text-buffer.coffee
+++ b/src/text-buffer.coffee
@@ -108,7 +108,7 @@ class TextBuffer
     @digestWhenLastPersisted = params?.digestWhenLastPersisted ? false
 
     @setPath(params.filePath) if params?.filePath
-    @load(true) if params?.load
+    @load() if params?.load
 
   # Called by {Serializable} mixin during deserialization.
   deserializeParams: (params) ->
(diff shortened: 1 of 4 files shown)


maxbrunsfeld:
@kevinsawicki Could you take a look at this one, since you worked on the file loading/watching logic? It's seeming like it's ok to just always use the serialized text when deserializing the buffer, but I might be missing something.

src/text-buffer.coffee
```diff
@@ -107,7 +107,6 @@ class TextBuffer
     @transactCallDepth = 0
     @digestWhenLastPersisted = params?.digestWhenLastPersisted ? false
     @modifiedWhenLastPersisted = params?.modifiedWhenLastPersisted ? false
```
kevinsawicki: I think `@modifiedWhenLastPersisted` can be removed now too, it appears unused

kevinsawicki:
> Could you take a look at this one, since you worked on the file loading/watching logic? 

This looks okay to me, I'm not 100% sure why these nuances were there previously but if the specs are green then I'm good with this. I did leave one minor comment about another piece of serialized state that could possibly be removed now as well since it is unused.

PR #102, opened 2015-09-25
Fix onDidChangeModified events after deserialization

maxbrunsfeld:
Refs https://github.com/atom/atom/issues/8849#issuecomment-143153421


maxbrunsfeld:
Oops, reopening this against the 6.x branch, since Atom stable is still on version 6.

PR #103, opened 2015-09-25
Use 1-dimensional marker index

nathansobo:
This PR swaps out the original 2D marker index with a dependency on the new [marker-index](https://github.com/atom/marker-index) library.

The 1D index is more efficient and less memory intensive, but using it here requires more translation between 2D row/column coordinates and character indices when using markers. We should investigate the overhead of that translation before merging this and possibly wait until we apply some planned optimizations to the mapping.

I was hoping this might be the solution to our marker population pauses when executing large searches, but though it improves things, on its own it isn't a big enough improvement. So I'm going to try a completely different approach that limits the overhead of find and replace highlights to the visible row range. I also plan on using the 1D marker-index in that code, so ultimately it would be good to use in the text buffer as well so we only have to maintain the code of a single marker index.


nathansobo:
Turns out it's more expensive to translate points to offsets and vice versa than to just store things in 2d to begin with.

PR #109, opened 2015-11-03
Use treap-based marker index

nathansobo:
This PR switches to using a new treap-based [marker-index](https://github.com/atom/marker-index) library, which is about 3x faster. Because the `dump` operation on this new index is incapable of omitting portions of the tree that don't contain markers on which we're maintaining history, this PR also makes a breaking API change to move the `maintainHistory` option to marker layers rather than individual markers.

The `maintainHistory` change in turn required an implementation change in Atom core in order to maintain selections in their own layer, necessitating a few more API enhancements such as the ability to copy a marker layer and inclusion of history-tracked marker layer snapshots in the history. Since deprecated APIs were in my way as I made these changes, I went ahead and removed deprecated APIs as well.


nathansobo:
Published as v8.0.0.

PR #112, opened 2015-11-17
Remove unused local variables.

kseo:
oldExtent and newExtent are not used in TextBuffer.applyChange method.


maxbrunsfeld:
Thanks for catching that @kseo.

PR #113, opened 2015-11-17
Replace the body of ::toDelta method with @getExent().

kseo:
::getExtent and ::toDelta do the exactly same thing.


maxbrunsfeld:
Thanks for noticing this. I believe we can just :fire: `toDelta`. And from my searching, it is not used in any package.

kseo:
Sure. If it is not used, we don't need to duplicate the method.

PR #114, opened 2015-11-20
Use native MarkerIndex implementation

nathansobo:
Having implemented a simpler binary-tree algorithm for the `MarkerIndex`, I wanted to see what we could gain by [dropping it to C++](https://github.com/atom/marker-index/pull/2). Tests of both implementations vary pretty dramatically between runs, but I'm seeing roughly the following when searching for `e` in jQuery after 6 runs of each version:

| Implementation | Typical Time | Worst Time | Typical Total GC Time
| ------ | ----------- | ---- | ---- |
| native | 900ms | 1070ms | 90ms
| JS | 1200ms | 1500ms | 180ms

So the JS version is definitely more variable, slower overall, and puts more pressure on the garbage collector. A 300-600ms improvement based on my tests. Keep in mind there's a bunch of other stuff going on in this code path, but I wanted to measure a real-world use case for this data structure anyway.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "emissary": "^1.0.0",
     "event-kit": "^1.0.2",
     "fs-plus": "^2.0.0",
-    "marker-index": "^2.1.1",
+    "marker-index": "^3.0.2",
     "pathwatcher": "^6.2.3",
     "serializable": "^1.0.3",
     "span-skip-list": "~0.2.0",
--- a/src/marker-layer.coffee
+++ b/src/marker-layer.coffee
@@ -46,7 +46,6 @@ class MarkerLayer
     @index = new MarkerIndex
     @markersById = {}
     @markersIdsWithChangeSubscriptions = new Set
-    @nextMarkerId = 0
     @destroyed = false
     @emitCreateMarkerEvents = false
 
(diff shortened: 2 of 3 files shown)


nathansobo:
@maxbrunsfeld Do you by chance know the incantation I need to give Travis to use clang to build my module instead of gcc?

nathansobo:
Seems like I'm having an issue with the C++11 standard library being available.

maxbrunsfeld:
Currently, in marker-index's `binding.gyp` file, the `-std=c++11` and `-stdlib=libc++` are only set for OSX. We need to set the linux equivalents.

nathansobo:
@maxbrunsfeld Oh of course, we made it conditional. Do you happen to have an example of the Linux equivalents? And Windows? Did you do this for tree-sitter?

maxbrunsfeld:
Maybe just this:

```python
'cflags_cc': [
        '-std=c++0x',
]
```

nathansobo:
Looks like you can just leave the library off. I should set up a marker-index travis build.

nathansobo:
:green_apple:

nathansobo:
Probably should set this up on AppVeyor to make sure we're :cool: on Windows.

PR #116, opened 2015-12-07
Assign a unique id to new TextBuffer instances

as-cii:
So that we can have a unique identifier to deserialize buffers that do not have a path (i.e. untitled buffers).

/cc: @atom/feedback @nathansobo


nathansobo:
:+1:

PR #118, opened 2015-12-08
Fix ::isModified for unsaved buffers

as-cii:
While trying to break https://github.com/atom/atom/pull/9968, I noticed that after instantiating a `TextBuffer` without passing `load: true` when the buffer hadn't a file path (i.e. which is what happens when deserializing an untitled editor) caused `TextBuffer::isModified` to alter its behavior, thereby always returning `false`.

I think there was a misplaced guard clause in the `::isModified` method and I moved it, changing two existing specs to clarify this behavior, which now seems consistent also with the documentation:

```
# Public: Determine if the in-memory contents of the buffer differ from its
# contents on disk.
#
# If the buffer is unsaved, always returns `true` unless the buffer is empty.
#
# Returns a {Boolean}.
```

/cc: @nathansobo @maxbrunsfeld for :eyes:

### DIFF
--- a/spec/text-buffer-spec.coffee
+++ b/spec/text-buffer-spec.coffee
@@ -1404,6 +1404,15 @@ describe "TextBuffer", ->
         buffer.setText('\n')
         expect(buffer.isModified()).toBeTruthy()
 
+    it "returns true when constructing a non-empty buffer with no path and not loading it", ->
+      buffer.destroy()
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
:+1:

nathansobo:
@as-cii You should have publish access on this module now, so I'll let you take care of upgrading.

PR #119, opened 2015-12-08
Make TextBuffer::setEncoding faster

as-cii:
Refs: https://github.com/atom/atom/issues/9979

Before this PR, when setting the encoding for a buffer, we reloaded the whole file and cleared the undo stack afterwards. This can be particularly slow as the file gets bigger, because calling `::reload` will actually change the buffer through a diff in order to keep track of the history. 

screen shot 2015-12-08 at 16 30 07

However, since we clear the undo history right after reloading the file, we can pass `true` to `::reload` in order to avoid computing the diff altogether. Specs-wise nothing should change, as the behavior should be 100% backwards compatible.

/cc: @nathansobo @maxbrunsfeld @atom/feedback

### DIFF
--- a/src/text-buffer.coffee
+++ b/src/text-buffer.coffee
@@ -404,9 +404,7 @@ class TextBuffer
       @emitter.emit 'did-change-encoding', encoding
 
       unless @isModified()
-        @updateCachedDiskContents true, =>
-          @reload()
-          @clearUndoStack()
+        @updateCachedDiskContents true, => @reload(true)
     else
       @emitter.emit 'did-change-encoding', encoding
 


maxbrunsfeld:
Great catch!

kevinsawicki:
I thought we cleared the undo stack because previous transactions from other encodings wouldn't work once the encoding was changed.

So if you made some edits, changed the encoding, and then ran undo, the file would be corrupted.

maxbrunsfeld:
Yeah, we still clear the undo stack by calling `@reload(true)`.

We were just doing unnecessary work before by calling `@reload()` with no arguments. That makes the reload operation undoable by computing a diff between the buffer and the disk contents.

kevinsawicki:
> Yeah, we still clear the undo stack by calling @reload(true).

Oh awesome :+1:

krial057:
Thank you for the fast fix! :)

PR #120, opened 2015-12-15
Terminate backwardsScan correctly when last chunk contains matches

### LINKED ISSUE #111: `backwardsScan` + `replace` produces incorrect results
Hi, 

In javascript, the classic string replace function works as follows:

    "foo \n bar \r\n baz".replace(/\r?\n/g, '\r\n')
    > "foo \r\n bar \r\n baz"

whereas with TextBuffer the result is as follows:

    buffer = new TextBuffer("foo \n bar \r\n baz")
    buffer.backwardsScan( /\r?\n/g, match => match.replace('\r\n') )
    > "foo \n bar \n baz"

(...even more interesting things happen when replacing `/$/g`)

I have found this issue as I was exploring why suddenly my `editorconfig` stopped working in sindresorhus/atom-editorconfig#54

Can someone please have a look at this or help me understand if my issue with `editorconfig` is related to this.

maxbrunsfeld:
Fixes #111

### DIFF
--- a/spec/text-buffer-spec.coffee
+++ b/spec/text-buffer-spec.coffee
@@ -2214,7 +2214,7 @@ describe "TextBuffer", ->
 
     describe "when called with a random range", ->
       it "returns the same results as ::scanInRange, but in the opposite order", ->
-        for i in [1...10]
+        for i in [1...50]
           seed = Date.now()
           random = new Random(seed)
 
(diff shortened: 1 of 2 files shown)


spec/text-buffer-spec.coffee
```diff
@@ -2232,12 +2232,31 @@ describe "TextBuffer", ->
...
+            expect(backwardMatches).toEqual(forwardMatches, "Seed: #{seed}")
+          else
+            referenceBuffer = new TextBuffer(text: buffer.getText())
+            referenceBuffer.scanInRange regex, range, ({matchText, replace}) ->
+              replace(matchText + '.')
+
+            buffer.backwardsScanInRange regex, range, ({matchText, replace}) ->
+              replace(matchText + '.')
+
+            expect(buffer.getText()).toBe(referenceBuffer.getText(), "Seed: #{seed}")
```
maxbrunsfeld: The issue wasn't actually specific to `replace`; The existing tests would occasionally catch it. But I thought I'd add randomized tests for `replace` while I was here anyway.

lierdakil:
Great! Thanks!

PR #126, opened 2016-01-26
Provide consolidated changes to ::onDidStopChanging callbacks

maxbrunsfeld:
I think we can use this in the `spell-check` package to avoid re-checking the entire file every time the user stops typing.

/cc @nathansobo were you going to rename `replacedExtent` to `oldExtent` and `replacementExtent` to `newExtent` inside of `atom-patch`? If so, we should probably do that before merging this.

/cc @as-cii maybe this could be useful in the `autocomplete-plus` [SymbolProvider](https://github.com/atom/autocomplete-plus/blob/db35d7a9ae16d2a86f605e63f02140e22b364b92/lib/symbol-provider.coffee#L55) also.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
   "dependencies": {
     "delegato": "^1.0.0",
     "atom-diff": "^2",
-    "atom-patch": "0.0.2",
+    "atom-patch": "0.0.3",
     "emissary": "^1.0.0",
     "event-kit": "^1.0.2",
     "fs-plus": "^2.0.0",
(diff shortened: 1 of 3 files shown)


nathansobo:
> were you going to rename replacedExtent to oldExtent and replacementExtent to newExtent inside of atom-patch?

Yeah, I still want to do that. I can make that change now.

nathansobo:
@maxbrunsfeld atom-patch 0.0.3 has the new API.

as-cii:
Great! Looks good to me after the rename. :zap:

maxbrunsfeld:
Do you think we should leave the argument undocumented for now?

nathansobo:
It's probably lower risk to use it ourselves a bit before committing to it. I feel good about it but I'm wrong enough to be cautious.

PR #127, opened 2016-01-29
Introduce TextBuffer.prototype.onDidChangeText

as-cii:
As discussed yesterday, this PR adds the `onDidChangeText` API and closes atom/atom#10610. It felt more important to work on this first, so that we can use it in autocomplete-plus right away.

Two concerns:

1. Maintaining two patches (one for `onDidChangeText` and one for `onDidStopChanging`) is not ideal, as pointed out yesterday, because splicing into them has some cost and I am not sure if/how we can reduce it by reusing one as a starting point for the other. That said, I feel like the splay tree solution that @nathansobo suggested could be enough to make splices fast (a brief profiling session showed that we spend most of the time in tree traversals).
2. I feel like we'd need to change `atom-patch` to support `oldText` in addition to `newText`. This can be addressed in a different PR, as it's not strictly related to `text-buffer`. Still, it seems like a solid addition to the API and would be useful for autocomplete-plus as well.

None of the above is blocking for this PR, though, and we can work directly in `atom-patch` to solve both issues.

/cc: @nathansobo and @maxbrunsfeld for :eyes:

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
   "dependencies": {
     "delegato": "^1.0.0",
     "atom-diff": "^2",
-    "atom-patch": "0.0.4",
+    "atom-patch": "0.0.6",
     "emissary": "^1.0.0",
     "event-kit": "^1.0.2",
     "fs-plus": "^2.0.0",
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
   "dependencies": {
     "delegato": "^1.0.0",
     "atom-diff": "^2",
-    "atom-patch": "0.0.6",
+    "atom-patch": "0.0.8",
     "emissary": "^1.0.0",
     "event-kit": "^1.0.2",
     "fs-plus": "^2.0.0",
--- a/spec/text-buffer-spec.coffee
+++ b/spec/text-buffer-spec.coffee
@@ -2333,22 +2333,19 @@ describe "TextBuffer", ->
       waitsFor ->
         buffer.loaded
 
-    it "notifies observers at the end of each transaction", ->
+    it "notifies observers after a transaction, an undo or a redo", ->
       textChanges = []
-      buffer.onDidChangeText (changes) -> textChanges.push(changes)
+      buffer.onDidChangeText (changes) -> textChanges.push(changes...)
 
       buffer.insert([0, 0], "abc")
       buffer.delete([[0, 0], [0, 1]])
-      expect(textChanges.length).toBe(2)
-      expect(textChanges[0]).toEqual([
+      expect(textChanges).toEqual([
         {
           start: {row: 0, column: 0},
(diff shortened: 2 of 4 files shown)


nathansobo:
One thing I didn't consider previously is the type of the points in the change objects. Other buffer APIs return `Point` instances, whereas the patch simply returns `{row, column}` objects. Should we iterate over the changes and convert the points for consistency? In retrospect I regret having an explicit `Point` class, but I'm not sure it's worth diverging from what has come before.

as-cii:
That's a good point, @nathansobo and I agree we should be consistent with the old interface; I extracted a `normalizePatchChanges` helper function to deal with this conversion.

as-cii:
I updated this PR with the new batching/rebalancing features introduced in atom/atom-patch#1. 

I think we can :shipit: after one last :eyes: and then go on merging atom/autocomplete-plus#659. :tada:

nathansobo:
There's a couple small things I'm not totally sure about...

* The method name: Are we comfortable with `onDidChangeText`? I'm wondering if we should name it `onDidTransact` so that the concept we're introducing is stable and more precisely describes this scenario. If we later wanted to go back and add other kinds of events, such as something that batches all changes in a tick, the more general name would remain available.

* The event object: Should it be an array of changes directly, or should we emit an object with a `changes` field? That would give us room to supply other information in the future gracefully. One idea that comes to mind is some kind of incrementing version number that could be used to interact with the history API. We might not need this, but does it make sense to leave room for it?

as-cii:
> The event object: Should it be an array of changes directly, or should we emit an object with a changes field? That would give us room to supply other information in the future gracefully. One idea that comes to mind is some kind of incrementing version number that could be used to interact with the history API. We might not need this, but does it make sense to leave room for it?

Good point, I recall that having an object has turned out to be handy in the past so I agree to use it here as well.

> The method name: Are we comfortable with onDidChangeText? I'm wondering if we should name it onDidTransact so that the concept we're introducing is stable and more precisely describes this scenario. If we later wanted to go back and add other kinds of events, such as something that batches all changes in a tick, the more general name would remain available.

I am not totally sure either: `onDidTransact` is definitely more precise here (as of the current implementation) but I am worried that it makes the aim of this method less clear. Namely, I wonder if people would expect that event to be triggered when e.g. `buffer.insert(...)` is called: internally the insertion is wrapped in a transaction, but as far as the user is concerned, they haven't explicitly transacted. What do you think?

I feel like these events will stay private for a while, so we'll probably have room for improving them as we make experiments.

as-cii:
I went ahead and merged this so that we can start using it in autocomplete-plus. If needed, we can refine the API during the experimentation phase. :+1:

maxbrunsfeld:
Great work @as-cii!

PR #129, opened 2016-02-06
Don’t serialize markers for layers with maintainHistory set to false

kuychaco:
This is to address a performance problem we're seeing in https://github.com/atom/atom/pull/10605

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "text-buffer",
-  "version": "8.2.1",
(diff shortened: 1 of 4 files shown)


kuychaco:
Decided to go a different direction and NOT serialize marker layers with `maintainHistory` set to false. This is to avoid memory leaks in packages that create new marker layers upon deserialization. See https://github.com/atom/text-buffer/pull/130

PR #132, opened 2016-02-12
Enable Babel And Upgrade To Jasmine 2.4.1

joefitzgerald:
This PR:

- [x] Updates Jasmine from 1.3.x to 2.4.1
- [x] Adds support for ES6/7 code by integrating babel into the test and compile steps
- [x] Adds support for linting ES6/7 code by integrating eslint
- [x] Reduces dependence on Grunt such that it is only used for the clean and atomdoc steps
- [x] Refactors spec-helper.coffee into JavaScript, and decomposes it slightly
- [x] Adds a babel helper to register babel for specs
- [x] Adds a coffee helper to register coffee for specs
- [x] Converts specs to use Jasmine 2.x syntax (e.g. `waits` / `waitsFor` >> `done`)
- [x] Removes the .pairs file, which seems to be unused
- [x] Fix: TextBuffer serialization when the buffer has a path when the serialized buffer had no unsaved changes loads the current contents of the file at the serialized path
- [x] Fix: TextBuffer serialization when the buffer has a path when the serialized buffer had unsaved changes when the disk contents were changed since serialization loads the disk contents instead of the previous unsaved state
- [x] Fix: TextBuffer modified status reports the modified status changing to true or false after the user changes buffer
- [x] Fix: TextBuffer modified status reports the modified status changing to false after a modified buffer is saved
- [x] Fix: TextBuffer modified status reports the modified status changing to false after a modified buffer is reloaded
- [x] Fix: TextBuffer modified status reports the modified status changing to false after a buffer to a non-existent file is saved
- [x] Fix: TextBuffer ::saveAs(path, {backup}) stops listening to events on previous path and begins listening to events on new path
- [x] Fix: TextBuffer ::onDidStopChanging(callback) notifies observers after a delay passes following changes

/cc @maxbrunsfeld


appveyor.yml
```diff
@@ -1,5 +1,5 @@
 environment:
-  nodejs_version: "4"
+  nodejs_version: "5"
```
joefitzgerald: The Travis build just specifies `node`, which translates to `5.x` currently.

package.json
```diff
@@ -4,10 +4,12 @@
   "description": "A container for large mutable strings with annotated regions",
   "main": "./lib/text-buffer",
   "scripts": {
-    "prepublish": "grunt clean lint coffee shell:update-atomdoc atomdoc",
-    "compile": "coffee --no-header --output lib --compile src",
-    "lint": "coffeelint -r src spec",
-    "test": "jasmine-focused --coffee --captureExceptions --forceexit spec",
+    "prepublish": "npm run clean && npm run compile && npm run lint && npm run atomdoc",
+    "atomdoc": "grunt shell:update-atomdoc atomdoc",
```
joefitzgerald: If there's a way to compile atomdoc outside grunt we could kill off grunt completely.

spec/helpers/set.js
```diff
@@ -0,0 +1,79 @@
...
+//   console.log(jasmine.getEnv())
+//   return jasmine
+//     .getEnv()
+//     .currentSpec
+//     .results()
+//     .getItems()
+//     .some((item) => {
+//       return !item.passed()
+//     })
+// }
```
joefitzgerald: This is problematic in Jasmine 2.x. Say the word and I will remove it completely.

spec/patch-spec.coffee
```diff
@@ -15,7 +15,6 @@ describe "Patch", ->
         expect(iterator.next()).toEqual {value, done: false}
         expect(iterator.getInputPosition()).toEqual inputPosition, "input position for hunk #{i}"
         expect(iterator.getOutputPosition()).toEqual outputPosition, "output position for hunk #{i}"
-        return if currentSpecFailed()
```
joefitzgerald: This seems to be just a debugging optimization, hence non-critical.

spec/patch-spec.coffee
```diff
@@ -496,9 +495,3 @@ describe "Patch", ->
           expectValidIterator(patch, iterator, position.traverse(newExtent))
           expectCorrectHunkMerging(patch)
           expectCorrectInternalNodes(patch.rootNode)
-
-          if currentSpecFailed()
```
joefitzgerald: This seems to be just a debugging optimization, hence non-critical.

spec/spec-helper.coffee
```diff
@@ -1,53 +0,0 @@
-require 'coffee-cache'
```
joefitzgerald: This file was converted to ES6.

.eslintrc
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "standard",
```
joshaber: We've been using plain ol' Standard on Atom itself and some internal packages. Any particular reason we need to extend it for this?

joefitzgerald: https://www.npmjs.com/package/eslint-plugin-promise << that's why.

joshaber: Huh, that _is_ nice!

spec/helpers/set.js
nathansobo: The main reason we need it is for randomized tests... we want them to terminate as soon as a single failure occurs rather than piling up more iterations. Is there a better approach to getting early termination in those tests?

spec/helpers/set.js
```diff
@@ -0,0 +1,79 @@
+'use babel'
+
+import {setEqual} from '../../src/set-helpers'
+
+Set.prototype.isEqual = (other) => { // eslint-disable-line no-extend-native
```
nathansobo: Can we avoid this?

spec/patch-spec.coffee
nathansobo: This is pretty important to me. There's no way to achieve it in Jasmine 2.0?

spec/helpers/set.js
nathansobo: I see, it was already there.

nathansobo:
One big issue with this is the ability to run tests in a browser with access to the Chrome dev tools. In other libraries such as [`atom-patch`](https://github.com/atom/atom-patch) I've used mocha and have an `npm run tdd` command that allows me to run the tests in an Electron window. This is critical when writing complex code where the interactive debugger and ability to write HTML make it much easier to reason.

Currently, we can just run `text-buffer` tests directly with Atom's Jasmine 1.3 runner. I'm fine switching to Jasmine 2.0, but we need some sort of story in place for the next time I want to run tests within Electron.

joefitzgerald:
> One big issue with this is the ability to run tests in a browser with access to the Chrome dev tools.

So, currently you can't do this, but this PR doesn't let you do it either. So it's not causing a regression, but this is a desirable additional capability?

spec/helpers/set.js
joefitzgerald: Avoid what? The eslint suppression, or the isEqual? Yes, this was converted verbatim, no additional things were added.

spec/patch-spec.coffee
joefitzgerald: Jasmine 2.3 (I think) introduced the ability to run the tests random, generally. It also has the ability to fail fast. So maybe we just turn that on?

maxbrunsfeld:
> So, currently you can't do this, but this PR doesn't let you do it either.

Well, we can currently use Atom's default test runner. With jasmine 2.0 though, we won't be able to, so we'll need some similar way of debugging. I think [this](https://github.com/electronjs/electron-jasmine) might be a good replacement.

spec/support/jasmine.json
```diff
@@ -0,0 +1,13 @@
...
+  "spec_files": [
+    "**/*.js",
+    "**/*.coffee"
+  ],
+  "helpers": [
+    "../node_modules/babel-core/register.js",
+    "helpers/**/*.js"
+  ],
+  "stopSpecOnExpectationFailure": false,
+  "random": false
```
joefitzgerald: @nathansobo Here's the `random` value that we could flip to `true`, and we could flip `stopSpecOnExpectationFailure` to `true` to stop immediately?

joefitzgerald:
> Well, we can currently use Atom's default test runner. With jasmine 2.0 though, we won't be able to, so we'll need some similar way of debugging. I think [this](https://github.com/electronjs/electron-jasmine) might be a good replacement.

Ahh, gotcha - `electron-jasmine` looks great for that!

spec/helpers/set.js
joefitzgerald: See my comment below about `random` and `stopSpecOnExpectationFailure `.

joefitzgerald:
Hmm...

```
$ node_modules/.bin/electron-jasmine ./spec
(electron) ipc module is deprecated. Use require("electron").ipcMain instead.
App threw an error when running [Error: Electron compilers not found but were requested to be loaded]
```

nathansobo:
Yikes. Using deprecated APIs. It would probably be pretty easy to hammer it into shape though. As long as I can run specs in Electron in some way, I'm on board with these improvements.

joefitzgerald:
Yeah it looks like it's doable to update electron-jasmine to work with the latest electron. I might need a bit of help with that; as I know relatively little about electron, but in doing a like poking around, it sounds like we might want to ditch the dependency on `electron-compile`, and update the dependency on `electron-prebuilt`, and otherwise use a similar approach to [`electron-mocha`](https://github.com/jprichardson/electron-mocha).

nathansobo:
Another important detail is that the original Electron runner for Mocha only ran in batch mode, launching tests in a hidden window and logging to the terminal. That's okay for some cases, but the real need in this case is to actually run and interact with the tests in Electron.

joefitzgerald:
So do you want me to flip `random` to `true` and `stopSpecOnExpectationFailure` @nathansobo? or is that just something you'd need to know how to flip when you're interactively working on the code?

nathansobo:
> So do you want me to flip random to true and stopSpecOnExpectationFailure @nathansobo? or is that just something you'd need to know how to flip when you're interactively working on the code?

Sorry, forgot to mention that. I think I'm cool flipping those flags when needed. It would be *nice* to do it just for that one test always, but it's not super critical.

joefitzgerald:
I'm guessing flipping random on will cause failures due to lifecycle issues in the specs. It might actually be worthwhile to open an issue to get to the point where it can be turned on by default?

nathansobo:
Yeah, that would be great, but not a priority to me at all. I just need the immediate failure for cases where I'm trying thousands of random combinations in fuzz tests.

joefitzgerald:
:shipit: ?

nathansobo:
Thanks @joefitzgerald. I have some other stuff to deal with before publishing this, but wanted to get it on master in case you wanted to build on top of it.

PR #136, opened 2016-03-03
Redesign History and Improve serialization performance

as-cii:
Depends on https://github.com/atom/atom-patch/pull/8.
Depends on https://github.com/atom/atom-patch/pull/7.

My main concern about the design yesterday was the inconsistency between single changes and Patches: a possible solution could have been to represent single changes as `Patch`es, but I was afraid of the possible overhead of splicing into a `Patch` instead of simply pushing an entry into an array.

After talking to @nathansobo yesterday and reasoning a bit more about this fine line between design and performance, I found out that we were already treating the result of `Patch.compose` as read-only (e.g. we didn't output a `Patch` but just the changes). For this reason, providing a bunch of read-only primitives that have the same interface as `Patch` but without having the same writing costs seemed very reasonable and straightforward. Also, now I feel like storing the `oldText` in `Patch` is actually nice design-wise because all the state is encoded in that data structure, and if we want to invert it we just need to call a method on it without referring to the current state of the buffer.

What's missing:

* [x] Delete entries from history by taking into account that a single `Patch` could contain many changes.
* [x] Should we provide backwards-compatibility with the previous `History` serialization format?
* [x] Serialize markers only when exiting the application (i.e. by passing `markerLayers: false` to `serialize`)

/cc: @nathansobo @maxbrunsfeld @kuychaco


src/text-buffer.coffee
```diff
@@ -120,33 +120,37 @@ class TextBuffer
...
-      markerLayers[layerId] = MarkerLayer.deserialize(this, layerState)
+      markerLayers[layerId] = MarkerLayer.deserialize(buffer, layerState)
     params.markerLayers = markerLayers
     params.defaultMarkerLayer = params.markerLayers[params.defaultMarkerLayerId]
-    params.history = History.deserialize(this, params.history)
+    params.history = History.deserialize(buffer, params.history)
     params.load = true if params.filePath
-    params
+    buffer.initialize(params)
+    buffer
```
as-cii: Removing the serializable mixin highlighted an interesting design problem, where some of our dependencies (e.g. `MarkerLayer` and `History` are cyclic). I think we might do something about it, but I didn't want to address it just yet and I simply worked it around by extracting an `initialize` method and calling it manually when deserializing. Thoughts? :thought_balloon:

src/text-buffer.coffee
```diff
@@ -65,8 +64,6 @@ class TextBuffer
   @Patch: require('./patch')
   @newlineRegex: newlineRegex
 
-  Serializable.includeInto(this)
```
as-cii: Here I have removed `Serializable` so that we are allowed to pass arbitrary parameters to the `TextBuffer.prototype.serialize` method, such as the one to exclude marker layers from being serialized.

PR #138, opened 2016-03-10
Stop using a forked version of `diff`

### LINKED ISSUE #137: Painfully slow buffer-update for large files
## Problem description
If a large file is open in an atom editor window and it is reloaded for whatever reason, it will block the entire editor. This will happen for a painfully long time.

Very often I will have a large JSON file open and for one reason or another it will be reloaded, typically caused by a fs change, and the entire browser will be blocked for what is often a minute or more.

Today I finally performed a profile of the issue and nailed down the cause.
Here is the result for the reload of a _relatively medium sized_ file (only ~6000 lines).
Profiling Result

### Short term
Give the user the option to diff the file or simply forego the ability to undo the reload **if the file is large**.
 - Reduces user frustration by eliminating unexpected behavior
 - Retains functionality _if_ one wants it
 - Easy to implement, shouldn't take more than a couple LOC :+1: 

### Long term
Somehow preserve undo-ability for massive files 

## Execution
Let me know if you like the sound of the short term solution, I'd be happy to implement it ASAP

50Wliu:
The only change in the forked version was merged into the main repo 2+ years ago.

Fixes #137

/cc @ArekSredzki, @kevinsawicki, @benogle

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
   },
   "dependencies": {
     "delegato": "^1.0.0",
-    "atom-diff": "^2",
+    "diff": "^2.2.1",
     "atom-patch": "0.2.0",
     "emissary": "^1.0.0",
     "event-kit": "^1.0.2",
--- a/spec/text-buffer-spec.coffee
+++ b/spec/text-buffer-spec.coffee
@@ -1183,15 +1183,15 @@ describe "TextBuffer", ->
         expect(calls).toBe(0)
 
       it "changes the in-memory contents of the buffer to match the new disk contents and notifies ::onDidChange observers", ->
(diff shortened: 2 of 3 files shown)


50Wliu:
Hmm.  Looks like this isn't going to be a clean switch.

nathansobo:
@ArekSredzki and @50Wliu, I appreciate your stewardship on this. Very glad we're off the forked version. Keep up the good work.

nathansobo:
Published as `text-buffer@8.4.2`. Thanks again!

ArekSredzki:
@nathansobo Thanks! :smile:

PR #139, opened 2016-03-10
Fixed compatability with new jsdiff version

ArekSredzki:
The output order for line removal & addition has changed in the new versions of jsdiff.
I have updated the app & tests accordingly.
Also added a clarifying comment so that no one else gets caught by why using the `count` property of the jsdiff output causes problems :)

### DIFF
--- a/spec/text-buffer-spec.coffee
+++ b/spec/text-buffer-spec.coffee
@@ -1183,15 +1183,15 @@ describe "TextBuffer", ->
         expect(calls).toBe(0)
 
       it "changes the in-memory contents of the buffer to match the new disk contents and notifies ::onDidChange observers", ->
-        expect(event1.oldRange).toEqual [[0, 0], [0, 0]]
-        expect(event1.newRange).toEqual [[0, 0], [0, 6]]
-        expect(event1.oldText).toBe ""
-        expect(event1.newText).toBe "second"
+        expect(event1.oldRange).toEqual [[0, 0], [0, 5]]
+        expect(event1.newRange).toEqual [[0, 0], [0, 0]]
+        expect(event1.oldText).toBe "first"
+        expect(event1.newText).toBe ""
 
-        expect(event2.oldRange).toEqual [[0, 6], [0, 11]]
-        expect(event2.newRange).toEqual [[0, 6], [0, 6]]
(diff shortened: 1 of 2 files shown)


ArekSredzki:
@50Wliu Please check this over and merge it accordingly

50Wliu:
@ArekSredzki Is this change backwards-compatible?  Why have the ranges changed in the specs?

As I'm not familiar with TextBuffer, I don't feel qualified to give this a full review.  @nathansobo it looks like you implemented this method.

ArekSredzki:
@50Wliu I'm not sure what you mean by backwards-compatible. It requires a different npm package entirely, so it doesn't matter.

As far as functionality, it's the same. The reason that the ranges have changes is that it used to add the new line and remove the old part, but now it removes the old one and then adds the new one.

50Wliu:
@ArekSredzki Yeah, I meant functionality-wise, so that's good to know.  Any reason why you switched around the order?  I'm guessing because that's how jsdiff operates?

ArekSredzki:
@50Wliu I switched the order solely because jsdiff change the order in it's output (I believe they did this when adding support for outputting structured patch files). The fact that they changed the order is what broke the text-buffer usage of the module.

50Wliu:
Awesome, thanks for clarifying things for me :).

ArekSredzki:
Happy to help!

ArekSredzki:
@nathansobo bump for review :)

nathansobo:
I'm going to be reviewing a bunch of PRs next week. I'll add it to the queue.

ArekSredzki:
@nathansobo Great, thanks!

nathansobo:
Taking a look now.

spec/text-buffer-spec.coffee
```diff
@@ -1183,15 +1183,15 @@ describe "TextBuffer", ->
         expect(calls).toBe(0)
 
       it "changes the in-memory contents of the buffer to match the new disk contents and notifies ::onDidChange observers", ->
-        expect(event1.oldRange).toEqual [[0, 0], [0, 0]]
```
nathansobo: It's actually nice that the order is delete-then-insert now. Makes more sense to me intuitively.

PR #141, opened 2016-03-11
Fix specs on machines that use CRLF line endings

50Wliu:
I don't know why Appveyor didn't fail on these specs.  Maybe it's configured to use LF line endings, or maybe it's just my laptop?  /cc @joefitzgerald 

At any rate, this PR fixes around 7-8 specs that were previously failing on my laptop.
```
> jasmine --captureExceptions --forceexit

Started
............................................................
............................................................
............................................................
............................................................
..............................................


286 specs, 0 failures
Finished in 10.983 seconds
```

Perhaps `os.EOL` isn't the best way to fix this.  I'm open to replacing all instances of `\r\n` to `\n` instead or using an even better way if there is one.

### DIFF
--- a/spec/text-buffer-spec.coffee
+++ b/spec/text-buffer-spec.coffee
@@ -1,5 +1,6 @@
 fs = require 'fs-plus'
 {join} = require 'path'
+os = require 'os'
 temp = require 'temp'
 {File} = require 'pathwatcher'
 Random = require 'random-seed'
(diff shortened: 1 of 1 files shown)


50Wliu:
...of course Appveyor is failing now.  I suspect it's my laptop then.

joefitzgerald:
This is a git issue. You need to set autocrlf settings to none, for just the fixtures directory.

PR #152, opened 2016-04-25
Deprecate individual Marker events

as-cii:
Depends on #149. The actual diff is [here](https://github.com/atom/text-buffer/compare/display-layers...deprecate-marker-events).

Subscribing and emitting events for every single marker is an expensive operation for a number of reasons, and its cost increases as the number of markers goes up: for instance, just emitting the event with no information becomes onerous when there are lots of markers and/or subscribers.

With this pull-request we want to make subscribing to individual marker event an anti-pattern and thus deprecate it. We have planned to do so by changing `MarkerLayer.prototype.onDidUpdate` in two ways:

1. We have introduced the `created`, `updated`, `touched` and `destroyed` sets in the event object, which respectively indicate the ids for markers that were created, manually updated, updated because of a textual change, and destroyed.
2. We have changed the way in which the event is scheduled. In particular, subscribers will now be notified at the end of the transaction wrapping the marker changes, or synchronously if no transaction was initiated.

Task list:

* [ ] Rebase on top of **master** after 🚢ping #149.

/cc: @nathansobo


aminya:
@as-cii @smashwilson 

Is this PR still relevant? I know you are not involved with Atom anymore, but I want to know if I should spend time on reviving this pull request.

PR #155, opened 2016-05-06
Improve behavior when files are changed/deleted after TextBuffer has been serialized

as-cii:
This pull-request aims at improving the experience when files get changed or deleted after serializing a `TextBuffer` instance. In particular, these scenarios have been changed:

1. When serializing a buffer, deleting the underlying file and then deserializing the buffer.
  1. **Before this pull-request.** Clear the text and the undo stack.
  1. **After this pull-request.** Keep the text at the moment of serialization, mark the buffer as modified.
1. When serializing a buffer with unsaved changes, changing the underlying file and then deserializing the buffer.
  1. **Before this pull-request.** Reload the file, and clear the undo stack.
  1. **After this pull-request.** Keep the text at the moment of serialization, mark the buffer as modified, mark the buffer as _in conflict_.
1. When serializing a buffer with no unsaved changes, changing the underlying file and then deserializing the buffer.
  1. **Before this pull-request.** Reload the file, and clear the undo stack.
  1. **After this pull-request.** Reload the file, and keep the undo stack.

This should result in a more robust experience, thereby preventing users from feeling like they've lost work, but still trying to update the file to the latest version on disk when there are no conflicts and no pending changes.

/cc: @atom/core

### DIFF
--- a/spec/text-buffer-spec.coffee
+++ b/spec/text-buffer-spec.coffee
@@ -1078,16 +1078,14 @@ describe "TextBuffer", ->
             buffer2.setText(previousText)
             expect(buffer2.isModified()).toBeFalsy()
 
-      describe "when the serialized buffer was unsaved and had no path", ->
-        it "restores the previous unsaved state of the buffer", ->
-          buffer.destroy()
+    describe "when the serialized buffer was unsaved and had no path", ->
+      it "restores the previous unsaved state of the buffer", ->
+        buffer = new TextBuffer()
+        buffer.setText("abc")
 
-          buffer = new TextBuffer()
-          buffer.setText("abc")
-
-          buffer2 = TextBuffer.deserialize(buffer.serialize())
-          expect(buffer2.getPath()).toBeUndefined()
-          expect(buffer2.getText()).toBe("abc")
+        buffer2 = TextBuffer.deserialize(buffer.serialize())
+        expect(buffer2.getPath()).toBeUndefined()
+        expect(buffer2.getText()).toBe("abc")
 
   describe "::getRange()", ->
     it "returns the range of the entire buffer text", ->
--- a/spec/text-buffer-spec.coffee
+++ b/spec/text-buffer-spec.coffee
@@ -1335,27 +1335,19 @@ describe "TextBuffer", ->
         expect(bufferToDelete.isModified()).toBeTruthy()
 
     describe "when the file is not modified", ->
-      [previousContents, modifiedEvents] = []
-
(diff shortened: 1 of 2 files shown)


lee-dohm:
When you're referring to "clearing" or "keeping" the history in the above, you're referring to the undo history?

What does marking the buffer as in conflict do?

I feel like item 2 with an Active Buffer should remain as before (or have an option), that Atom should destroy the buffer. In my experience, this is most people's expectations. They find having to close the tab and choose not to save the contents after they've specifically deleted the file to be tiresome (especially if they deleted the file via the tree-view).

as-cii:
> What does marking the buffer as in conflict do?

We don't actually use it in Atom, but it's a state that we support in this library. It basically means that the changes you have made refer to an older version of the file on disk.

> When you're referring to "clearing" or "keeping" the history in the above, you're referring to the undo history?

Yup.

> I feel like item 2 with an Active Buffer should remain as before (or have an option), that Atom should destroy the buffer. In my experience, this is most people's expectations. They find having to close the tab and choose not to save the contents after they've specifically deleted the file to be tiresome (especially if they deleted the file via the tree-view).

👍 for the tree-view scenario, that's something our users should not think about. However, I am a bit concerned about the consistency of this behavior: in particular, after the buffer has been serialized without unsaved changes and the underlying file gets deleted in an external process, with this pull-request we would load the text stored in the serialized state after deserialization, and mark the buffer as modified. Do you think we should destroy the buffer in those scenarios too? 💭

as-cii:
> 👍 for the tree-view scenario, that's something our users should not think about. However, I am a bit concerned about the consistency of this behavior: in particular, after the buffer has been serialized without unsaved changes and the underlying file gets deleted in an external process, with this pull-request we would load the text stored in the serialized state upon deserialization, marking the buffer as modified. Do you think we should destroy the buffer in those scenarios too? 💭

For posterity: upon further discussion in chat, @lee-dohm and I agreed that whenever Atom is reloaded (and the buffer gets deserialized) but the file got deleted, we should always restore the previous contents of the buffer, otherwise users might feel like they've lost work.

as-cii:
@lee-dohm: I'd ❤️ if you could give this a try so that we can go on and 🚢 it. Thanks! 🙇

lee-dohm:
Yep, I'll take a look at this later today.

as-cii:
Closing this as now we have a better understanding of what could be causing the files to be wiped out. More on this soon in a different pull-request on Atom.

PR #162, opened 2016-07-29
Restore assignment of properties in MarkerLayer.prototype.markPosition

nathansobo:
@lierdakil is there anything else we broke that needs to be restored? Seems like we were still working correctly in `markRange`.

### DIFF
--- a/spec/marker-spec.coffee
+++ b/spec/marker-spec.coffee
@@ -73,6 +73,10 @@ describe "Marker", ->
         expect(buffer.findMarkers({})).toEqual [marker1]
         expect(buffer.getMarkers()).toEqual [marker1]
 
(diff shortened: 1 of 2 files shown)


lierdakil:
Yes, `markRange` seems to working pretty much as expected. I can't say I have tested this rigorously, or that I'm familiar with Atom's code base, but I think only `markPosition` was entirely broken. Thanks again.

nathansobo:
Yeah just wanted to make sure there wasn't anything else you had encountered that was definitely broken.

nathansobo:
This should go out on Monday on all channels. I'd have it out sooner but we're transitioning some aspects of our build and can't really do a hotfix right now.

lierdakil:
No problem, Monday sounds good.

PR #169, opened 2016-08-10
Copy foldCharacter and atomicSoftTabs in DisplayLayer.prototype.copy()

as-cii:
When adding the `atomicSoftTabs` and `foldCharacter` parameters, we forgot to do so also in `.prototype.copy()`. This could lead to unfortunate behaviors like splitting a pane and have the cursor behave differently depending on which pane the user is working on.

With this pull-request we add the missing settings to `copy()` along with a test that makes sure we don't regress in the future.

### DIFF
--- a/src/display-layer.coffee
+++ b/src/display-layer.coffee
@@ -62,7 +62,6 @@ class DisplayLayer
 
   constructor: (@id, @buffer, settings={}) ->
     @displayMarkerLayersById = {}
-    @textDecorationLayer = null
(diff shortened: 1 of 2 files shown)


src/display-layer.coffee
```diff
@@ -62,7 +62,6 @@ class DisplayLayer
 
   constructor: (@id, @buffer, settings={}) ->
     @displayMarkerLayersById = {}
-    @textDecorationLayer = null
```
as-cii: It should be okay to 🔥 this because we are assigning that ivar to an `EmptyDecorationLayer` on line 75.

PR #170, opened 2016-08-10
Prevent DisplayLayer state from becoming inconsistent with nested did-change event handlers

as-cii:
Refs: atom/atom#12085

After #168, we have observed that sometimes change events are emitted in a order we don't expect. A similar situation could be reproduced by creating a display layer, adding a `onDidChange` handler in `TextBuffer` and then creating a second display layer by calling `copy()` on the first one. If the `onDidChange` handler applies some modification to the text, as soon as an edit is made on the buffer the second display layer will receive change events in a reversed order and, therefore, will splice invalid regions into the display index.

With this pull-request we want to ensure that decoration layers and display layers are always updated in lockstep, thereby preventing them from having an inconsistent state with respect to each other.

Another positive aspect of this change is that it should shield against possible errors happening in third-party `did-change` event handlers that get registered before the display layer event handler. In facts, such errors might cause the change event to not trigger sometimes on the display layer, and thus to throw errors on future updates to the spatial lines index.

For these reasons, we are introducing a `registerTextDecorationLayer` method on `TextBuffer` that needs to be called by decoration layers when they are initialized. Doing so will enforce that change events are emitted first on decoration layers, then on display layers, then on `TextBuffer.onDidChange` subscribers, and finally on `DisplayLayer.onDidChangeSync` subscribers. The `registerTextDecorationLayer` method returns a `Disposable`, which should be called when the decoration layer is destroyed in order to stop receiving updates.

An alternative implementation might have computed all the decoration layers based on the registered display layers on a particular `TextBuffer` instance. This, however, would have been somewhat inconvenient because it would have forced users of `TokenizedBuffer` to always create a `DisplayLayer`, introducing some unnecessary coupling between the two objects.

/cc: @nathansobo

### DIFF
--- a/spec/helpers/test-decoration-layer.coffee
+++ b/spec/helpers/test-decoration-layer.coffee
@@ -19,7 +19,7 @@ class TestDecorationLayer
       @markerIndex.insert(markerId, Point.fromObject(rangeStart), Point.fromObject(rangeEnd))
       @tagsByMarkerId[markerId] = tag
 
-    @buffer?.preemptDidChange(@bufferDidChange.bind(this))
+    @buffer?.registerTextDecorationLayer(this)
 
   buildIterator: ->
     new TestDecorationLayerIterator(this)
--- a/spec/text-buffer-spec.coffee
+++ b/spec/text-buffer-spec.coffee
@@ -124,17 +124,33 @@ describe "TextBuffer", ->
         eventId: 1
       }]
 
-    it "notifies ::onDidChange observers with the relevant details after a change", ->
-      changes = []
-      buffer.onDidChange (change) -> changes.push(change)
+    it "notifies, in order, registered decoration layers, display layers and ::onDidChange observers with the relevant details after a change", ->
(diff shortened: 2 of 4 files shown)


spec/helpers/test-decoration-layer.coffee
```diff
@@ -19,7 +19,7 @@ class TestDecorationLayer
       @markerIndex.insert(markerId, Point.fromObject(rangeStart), Point.fromObject(rangeEnd))
       @tagsByMarkerId[markerId] = tag
 
-    @buffer?.preemptDidChange(@bufferDidChange.bind(this))
+    @buffer?.registerTextDecorationLayer(this)
```
as-cii: Alternatively we might register just the handler function instead of the whole text decoration layer object. Still, it feels somewhat convenient to have access to the entire object and reads better than `registerTextDecorationLayerChangeHandler` or `addTextDecorationLayerChangeHandler`.

PR #171, opened 2016-08-12
Compute screen lines lazily after a reset

maxbrunsfeld:
This is one solution to the problem of resetting a TextEditor's display layer repeatedly when configuring it from the outside (https://github.com/atom/atom/pull/12125).

I believe this brings file-opening performance on my Atom PR back to parity with master. I'm not sure that I love it, but opening for your consideration @as-cii @nathansobo.

### DIFF
--- a/spec/display-layer-spec.coffee
+++ b/spec/display-layer-spec.coffee
@@ -6,6 +6,7 @@ Range = require '../src/range'
 WORDS = require './helpers/words'
 SAMPLE_TEXT = require './helpers/sample-text'
 TestDecorationLayer = require './helpers/test-decoration-layer'
+DisplayIndex = require 'display-index'
 
 describe "DisplayLayer", ->
   beforeEach ->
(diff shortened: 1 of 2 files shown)


src/display-layer.coffee
```diff
@@ -1212,6 +1223,7 @@ class DisplayLayer
     }
 
   getScreenLineCount: ->
+    @recomputeScreenLinesIfNeeded()
```
as-cii: Should we add this to methods like `getRightmostScreenPosition` and `lineLengthForScreenRow` as well?

src/display-layer.coffee
```diff
@@ -126,6 +127,9 @@ class DisplayLayer
       "\r\n": @invisibles.cr + @invisibles.eol
     }
 
+  recomputeScreenLinesIfNeeded: ->
```
as-cii: What about renaming this to `recomputeAllSpatialScreenLinesIfNeeded()`? In this file we usually refer to "screen lines" when talking about the synthesized lines that contain both spatial information and decoration tags.

spec/display-layer-spec.coffee
```diff
@@ -33,6 +34,39 @@ describe "DisplayLayer", ->
       expect(displayLayer2.ratioForCharacter).toBe(displayLayer1.ratioForCharacter)
       expect(displayLayer2.isWrapBoundary).toBe(displayLayer1.isWrapBoundary)
 
+  it "does not recompute the display layer's screen lines until they are needed", ->
+    spliceSpy = spyOn(DisplayIndex.prototype, 'splice').and.callThrough()
+    buffer = new TextBuffer(text: '\tif (a) { b; } else { c; }')
+
+    # construction does not force screen line computation
+    displayLayer = buffer.addDisplayLayer(tabLength: 4)
+    expect(spliceSpy.calls.count()).toBe(0)
```
as-cii: I think I am 👍 on spying on `splice`. Just as an alternative, we might consider querying the state of the `DisplayIndex` and avoid the spy:

```js
expect(displayLayer.displayIndex.getScreenLineCount()).toBe(1)
```

src/display-layer.coffee
```diff
@@ -126,6 +127,9 @@ class DisplayLayer
...
+    return unless @propertiesHaveChanged
+    @propertiesHaveChanged = false
```
as-cii: Super minor stylistic comment here. 🎨 What do you think about changing this into:

```coffee
if @propertiesHaveChanged
  @propertiesHaveChanged = false
  # etc. etc.
```

Feel free to ignore this if you don't think it helps readability. 👍

as-cii:
Left a couple of comments, but overall I think  I am 👍 with this approach. We are going to need something similar anyways if we want to batch buffer changes in the future, so it feels reasonable to me to go down this path.

nathansobo:
What about the problem of delaying the change event after calling `reset`?

maxbrunsfeld:
I think I'm putting this on 🍧  in favor of a different solution in Atom.

spec/display-layer-spec.coffee
maxbrunsfeld: Good point; that's a much better way of writing the assertion.

PR #175, opened 2016-10-06
Avoid computing spatial screen lines until they are needed

maxbrunsfeld:
Previously, creating a `DisplayLayer` would take an amount of time proportional to the size of the underlying buffer, because all of the screen lines were computed immediately.

Now, the `DisplayLayer` waits to compute *any* screen lines until they are needed for some query (e.g. `getScreenLines` or `translateBufferPosition`) or to interpret a buffer change event. This way, `TextBuffer.addDisplayLayer()` is a constant-time operation.

There is also a new method, `DisplayLayer.doBackgroundWork(deadline)`, which can be used to compute additional screen lines in advance of any query that would demand them.

* [x] Defer computation of spatial screen lines until they are needed for individual queries.
* [x] Update randomized tests to make sure that they cover the case where the buffer changes in a region for which spatial screen lines have not yet been computed.
* [x] Add a new public method `DisplayLayer::doBackgroundWork(deadline)` that the TextEditor can call in an idle callback to compute additional spatial screen lines ahead of time.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "text-buffer",
-  "version": "9.3.1-3",
+  "version": "9.3.1-4",
   "description": "A container for large mutable strings with annotated regions",
   "main": "./lib/text-buffer",
   "scripts": {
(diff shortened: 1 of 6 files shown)


as-cii:
Looks good to me! Great work here, @maxbrunsfeld! ⚡️

PR #176, opened 2016-10-17
Atomic soft tabs bug

N0r1x:
Bug fix for https://github.com/atom/atom/issues/12884 that I traced back to a typing error (unless it's intended to work like this, which would confuse me) in this module. There isn't really a lot else to say about it, except that I was slightly confused on how the particular code in the display-layer file works, the cross reference from the TokenIterator with this file serving as it's producer (of tokens) and consumer was slightly unexpected. But you could probably chalk most of that onto my inexperience in JavaScript as well.

EDIT:typos


50Wliu:
Thanks for the contribution @N0r1x, and sorry that we didn't get to review this.  It looks like this bug has been fixed in the interim.

PR #178, opened 2016-10-29
Don't destroy buffer when underlying file is deleted.

iolsen:
WIP toward fixing https://github.com/atom/tabs/issues/306.

I actually feel the existing behavior is crazytown and the people complaining in the issue above are 100% right.  IMHO there shouldn't even be an option to maintain this anti-social, silently destructive behavior. 😜 

But I see there were actually quite a few people advocating for it in https://github.com/atom/tabs/issues/160 so, fine. Option it is.

However I do feel strongly that reverting the default is The Right Thing because the current behavior can cause lost data in a totally unexpected way. So people who want their tabs silently closing, as they do now, will have to go change an option.

With this change a buffer whose underlying file has been deleted is essentially treated as a new, unsaved file:
- When opened in the scope of a project its state will be saved until it's explicitly saved to a real file.
- When opened without a project (in tandem with the fix in https://github.com/atom/atom/pull/12922) there will be a prompt to save or discard.
- When closing the tab there will be a prompt to save or discard.

Still TODO:
- [ ] Fix the explicit file delete command _within Atom_ to close the buffer as it does before this change (atom/tree-view#1026)
- [ ] Add a setting to auto-close on delete (atom/atom#13657)
- [ ] Tests


vjeux:
That would be super useful! @kyldvs made it work as an init.js script while this is getting merged in and released.

```js
let updatedPrototype = false;
atom.workspace.observeTextEditors(function (editor) {
  if (updatedPrototype) {
    return;
  }

  const TextBuffer = editor.buffer.constructor;

  let inSubscribeToFile = false;
  const subscribeToFile = TextBuffer.prototype.subscribeToFile;
  TextBuffer.prototype.subscribeToFile = function () {
    inSubscribeToFile = true;
    const ret = subscribeToFile.apply(this, arguments);
    inSubscribeToFile = false;
    return ret;
  };

  const File = editor.buffer.file.constructor;

  let inOnDidDeleteCallback = false;
  const onDidDelete = File.prototype.onDidDelete;
  File.prototype.onDidDelete = function () {
    let fn = arguments[0]; // Should be a function.
    if (inSubscribeToFile) {
      const ogFn = fn;
      fn = function () {
        inOnDidDeleteCallback = true;
        const ret = ogFn.apply(this, arguments);
        inOnDidDeleteCallback = false;
        return ret;
      };
    }
    return onDidDelete.apply(this, [fn]);
  };

  const destroy = TextBuffer.prototype.destroy;
  TextBuffer.prototype.destroy = function() {
    if (inOnDidDeleteCallback) {
      // Don't destroy it.
    } else {
      destroy.apply(this, arguments);
    }
  }

  updatedPrototype = true;
});
```

iolsen:
@nathansobo I don't love this, but it works so I went ahead and pushed it anyway for discussion. See also https://github.com/atom/atom/pull/13657.

src/text-buffer.coffee
```diff
@@ -131,6 +133,9 @@ class TextBuffer
     preferredLineEnding: @preferredLineEnding
     nextMarkerId: @nextMarkerId
 
+  setConfigCallbacks: (shouldDestroyBufferOnFileDelete) ->
```
nathansobo: What about using a constructor parameter named `shouldDestroyOnFileDelete` instead?

src/text-buffer.coffee
```diff
@@ -79,7 +82,7 @@ class TextBuffer
     @transactCallDepth = 0
     @digestWhenLastPersisted = params?.digestWhenLastPersisted ? false
 
-    @shouldDestroyBufferOnFileDelete = -> false
```
nathansobo: 👍

nathansobo:
:zap:

PR #181, opened 2016-11-03
Use a C++ implementation of BufferOffsetIndex

as-cii:
Supersedes and closes #177.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,18 +2,10 @@ language: node_js
 
 sudo: false
 
-addons:
-  apt:
-    sources:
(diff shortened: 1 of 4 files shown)


Swatinem:
Can you provide some insight into why you went with C++ rather than rust in the end?

as-cii:
> Can you provide some insight into why you went with C++ rather than rust in the end?

We really enjoyed using Rust to implement that index but we also felt like the tooling to integrate it with Node was still very young and not ready for production. So we decided to stick with C++, as it is a first-class citizen in the Node ecosystem.

PR #182, opened 2016-11-08
add new method getUnresolvedPath, update getPath to call realpath

dirk-thomas:
This patch is related to atom/atom#10856 which aims to allow passing non-realpath through the API (from `TextBuffer` to `Project`).

The new method `getUnresolvedPath` returns the unmodified path of the file. The existing `getPath` is being modified to continue returning a realpath even if the text buffer was constructed with a non-realpath. This way existing users of the API won't notice the change and can opt-in to the new method.

I haven't added a specific spec since I notices the one spec which was already working around the realpath problem by effectively ignoring the difference. I changed that spec to use the new API which should hopefully make it continue to pass (while actually checking the file path for equality).


dirk-thomas:
Could the failing AppVeyor build be unrelated to this patch?

dirk-thomas:
@maxbrunsfeld Is there anything I can do regarding the failing AppVeyor?

dirk-thomas:
Anyone?

50Wliu:
Try merging from master.

dirk-thomas:
Thank you for the hint that the problem has been fixed on master in the meantime. I have rebased the patch and it passed now.

dirk-thomas:
The goal is reached :+1: So all good :wink: Thank you.

PR #185, opened 2016-12-14
Optimize DisplayLayer's memory usage by storing transformations in a Patch

maxbrunsfeld:
#### The Problem
A major issue with Atom's current handling of large files is that the `DisplayLayer` uses so much memory that v8 can run out of heap space, causing a hard crash.

The reason that the `DisplayLayer` uses so much memory is that it stores all of the transformations that are applied to the text when it is displayed. These transformations include soft-wraps, folds, hard-tab expansion, atomic soft-tab clipping, indent guides, and invisible-character substitution. They are stored in a data structure called the [DisplayIndex](https://github.com/atom/display-index), which uses a binary tree to allow content to be accessed efficiently in terms of both buffer or screen coordinates.

#### The Optimization

Translating between arbitrary buffer and screen coordinates is an important responsibility of the DisplayLayer, but many of the transformations listed above do not affect position translation. The only transformations that *do* affect position translation are soft-wraps, folds, and hard-tabs.

This PR takes advantage of that affect by avoiding storing any of the other transformations, and computing them only as needed, when rendering lines of text. This means that the data *stored* by the DisplayLayer is now *sparse* with respect to screen lines; nothing needs to be stored for screen lines without folds, soft-wraps or hard tabs.

We now store the `DisplayLayer`'s soft-wraps and folds in a [`Patch`](https://github.com/atom/atom-patch) - a data structure that we already use elsewhere in this library, for representing groups of changes in the buffer's [`History`](https://github.com/atom/text-buffer/blob/08eb0ec1326919e1600b70f83bad4d4f7f114fc8/src/history.coffee), and for aggregating changes to pass to `onDidChangeText` callbacks.

As part of this effort, we have [rewritten](https://github.com/atom/atom-patch/pull/12) the Patch in C++ and with a more compact representation, in order to further reduce Atom's memory consumption when editing large files.


kylehotchkiss:
Came here via the release notes - does this improve file open time when syntax highlighting is applied as well or does that happen at another layer of the application?

maxbrunsfeld:
The main improvement is a dramatically lower memory consumption, so that you can open large files without crashing. File open time should improve across the board as well. This change is not specifically about syntax highlighting though.

kylehotchkiss:
@maxbrunsfeld just gave it a try on a 40mb / 1mil line JSON file and it works great. Thank you for your hard work!!!

PR #187, opened 2016-12-16
Optimize scan when using single-line regexps

maxbrunsfeld:
Most of the time when searching a buffer, the search regex can only match within a single line. This is the case whenever the regex doesn't contain an explicit line ending character (`\n` or `\r`) or a negated character class (e.g. `[^\w]`). In this situation, we can avoid two inefficient aspects of the current search algorithm:

1. We can avoid joining the buffer's text into a single string to pass to `Regex.exec` and then translating the resulting match positions back into 2-dimensional buffer ranges with the `BufferOffsetIndex`. Instead we can search each line individually, keeping track of the row as we go, so that we can provide the ranges with no cost.

2. When searching backwards, there's no need for the back-off algorithm that we currently use to avoid searching from the beginning of the text. We can simply iterate through the lines in reverse.

/cc @nathansobo

### DIFF
--- a/spec/display-layer-spec.js
+++ b/spec/display-layer-spec.js
@@ -3,8 +3,7 @@ const dedent = require('dedent')
 const TextBuffer = require('../src/text-buffer')
 const Point = require('../src/point')
 const Range = require('../src/range')
-
-const WORDS = require('./helpers/words')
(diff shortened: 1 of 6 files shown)


nathansobo:
This looks great.

PR #189, opened 2016-12-22
Don't yield empty matches at end of scanned range

nathansobo:
This behavior is required for Atom's `TextEditor` tests to pass the new line-based scanning introduced in #187. Basically, using a regex with a `*` was producing an extra empty match at the end of the scanned range in some cases.

/cc @maxbrunsfeld


maxbrunsfeld:
😓 Ah thanks for fixing that!

PR #190, opened 2016-12-28
Handle empty final row in TextBuffer.scanInRange

smashwilson:
If the final row of a range passed to `scanInRange` is empty, then the `break` statement shouldn't trigger on a match at the range's end, because a zero-length match _is_ a valid and expected result.

Manifesting in Atom as atom/atom#13489.


maxbrunsfeld:
Great explanation. Thanks for the fix @smashwilson!

PR #195, opened 2017-01-31
add context lines to matches

dirk-thomas:
This is in support of atom/find-and-replace#847.

The PR provide similar functionality as atom/scandal#41. It adds an optional `options` argument to the `scan` functions. The options can contain `lineCountBefore` and `lineCountAfter`. The return results now contain additionally `linesBefore` and `linesAfter`.

The previous signatures should continue to work without any need to change existing code.

### DIFF
--- a/src/match-iterator.coffee
+++ b/src/match-iterator.coffee
@@ -1,5 +1,19 @@
+_ = require 'underscore-plus'
 Point = require './point'
 Range = require './range'
+
+addContextLinesToCallbackArgument = (argument, options) ->
+  argument.leadingContextLines = []
+  row = Math.max(0, argument.range.start.row - (options.leadingContextLineCount or 0))
+  while row < argument.range.start.row
+    argument.leadingContextLines.push(argument.buffer.lineForRow(row))
+    row += 1
+
+  argument.trailingContextLines = []
+  for i in [0...(options.trailingContextLineCount or 0)]
+    row = argument.range.start.row + i + 1
+    break if row >= argument.buffer.getLineCount()
+    argument.trailingContextLines.push(argument.buffer.lineForRow(row))
 
 class SingleLineSearchCallbackArgument
   lineTextOffset: 0
@@ -16,9 +30,10 @@ class SingleLineSearchCallbackArgument
   Object.defineProperty @::, 'lineText',
     get: -> @buffer.lineForRow(@row)
 
-  constructor: (@buffer, @row, @match, @lineOffset) ->
+  constructor: (@buffer, @row, @match, @lineOffset, options={}) ->
     @stopped = false
     @matchText = @match[0]
+    addContextLinesToCallbackArgument(this, options)
 
   replace: (text) =>
     @replacementText = text
@@ -27,7 +42,7 @@ class SingleLineSearchCallbackArgument
   stop: => @stopped = true
 
 class ForwardsSingleLine
-  constructor: (@buffer, @regex, @range) ->
+  constructor: (@buffer, @regex, @range, @options={}) ->
 
   iterate: (callback, global) ->
     row = @range.start.row
@@ -37,7 +52,7 @@ class ForwardsSingleLine
 
     while row < @range.end.row
       if match = @regex.exec(line)
-        argument = new SingleLineSearchCallbackArgument(@buffer, row, match, lineOffset)
+        argument = new SingleLineSearchCallbackArgument(@buffer, row, match, lineOffset, @options)
         callback(argument)
         return if argument.stopped or not global
         if argument.replacementText?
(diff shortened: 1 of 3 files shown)


dirk-thomas:
Since the beta of 1.15 was just released it would great if this extended API could be merged to become available in the next beta (1.16).

src/text-buffer.coffee
```diff
@@ -1084,43 +1084,77 @@ class TextBuffer
   # {::backwardsScan} to avoid tripping over your own changes.
   #
   # * `regex` A {RegExp} to search for.
+  # * `options` (optional) {Object}
+  #   * `lineCountBefore` {Number} default `0`; The number of lines before the
+  #      matched line to include in the results object.
+  #   * `lineCountAfter` {Number} default `0`; The number of lines after the
+  #      matched line to include in the results object.
```
maxbrunsfeld: After talking with @nathansobo about this new API, we decided that the following names would tell the story better:

* `lineCountBefore` -> `leadingContextLineCount`
* `lineCountAfter` -> `trailingContextLineCount`
* `linesBefore` -> `leadingContextLines`
* `linesAfter` -> `trailingContextLines`

Could you make those changes to the field names? I'll update `scandal` to match those names.

src/text-buffer.coffee
```diff
@@ -1138,7 +1172,22 @@ class TextBuffer
...
+        break if row < 0
+        argument.linesBefore.unshift(@lineForRow(row))
+
+      argument.linesAfter = []
+      for i in [0...(options.lineCountAfter or 0)]
+        row = argument.range.start.row + i + 1
+        break if row >= @getLineCount()
+        argument.linesAfter.push(@lineForRow(row))
+
+      callback(argument)
```
maxbrunsfeld: Since `scanInRange` is called a lot, I'd like to avoid creating this wrapper closure. Could you move this functionality to the `SingleLineSearchCallbackArgument` and `MultiLineSearchCallbackArgument` classes?

I think that `iterator.iterate` could take `leadingContextLineCount` and `trailingContextLineCount` as two new parameters, and could pass them to the callback arguments' constructors.

src/text-buffer.coffee
```diff
@@ -1138,7 +1172,22 @@ class TextBuffer
       else
         iterator = new MatchIterator.ForwardsMultiLine(this, regex, range)
 
-    iterator.iterate(callback, global)
+    callback_with_context = (argument) =>
+      argument.linesBefore = []
+      for i in [0...(options.lineCountBefore or 0)]
+        row = argument.range.start.row - i - 1
+        break if row < 0
+        argument.linesBefore.unshift(@lineForRow(row))
```
maxbrunsfeld: Since `unshift` has to move every element in the array, I think it'd be slightly more direct and fast to iterate through the buffer's lines in the *forwards* direction. Maybe something like:

```coffee
row = Math.max(0, range.start.row - lineCountBefore)
while row < range.start.row
  argument.linesBefore.push(@lineForRow(row))
```

dirk-thomas:
I updated the code based on your feedback. I moved the logic to populate the context lines [here](https://github.com/atom/text-buffer/pull/195/files#diff-de8abe3e671124282519e9b9f62c4364R5) so that both `CallbackArgument` classes can use it.

One thing really confusing: Travis shows up "green" even if some tests are failing. That is probably something which should be addressed outside of this PR.

50Wliu:
All tests seem to be passing though?
```
365 specs, 0 failures
Finished in 15.361 seconds
```

dirk-thomas:
Yeah, now they are. Before I had some commits which failed some tests but Travis was still green.

I made an example PR to demonstrate the problem. Please see #222.

src/match-iterator.coffee
```diff
@@ -166,7 +182,11 @@ class ForwardsMultiLine
       match
 
 class BackwardsMultiLine
-  constructor: (@buffer, @regex, range, @chunkSize) ->
+  constructor: (@buffer, @regex, range, @options={}, @chunkSize) ->
+    if _.isNumber(@options)
+      @chunkSize = @options
+      @options = {}
```
maxbrunsfeld: Why is a conditional needed here? Don't we always pass the `chunkSize` argument? I'd think that we could make `options` the last argument.

dirk-thomas: The three other classes don't have a `chunkSize` argument. I tried to keep the order of the common arguments the same (therefore `options` after `range`). The conditional is there to not break compatibility if any other code is calling this API.

maxbrunsfeld: I see what you're saying; thanks for being careful about compatibility. In this case though, these classes are private and their constructors don't have to have the same signature.

dirk-thomas: Ok, no problem. I updated the patch accordingly.

maxbrunsfeld:
Thanks @dirk-thomas; Great work!

nathansobo:
⚡️ ⚡️ ⚡️

PR #201, opened 2017-02-09
Populate DisplayLayer's spatial index before destroying folds

maxbrunsfeld:
When a DisplayLayer is created with a non-empty folds marker layer (e.g. when reloading an Atom window after creating folds), it previously could get into an inconsistent state when those folds were destroyed. This fixes that.


Ben3eeE:
Sounds like https://github.com/atom/atom/issues/13258#issuecomment-261632556?

maxbrunsfeld:
> Sounds like atom/atom#13258 (comment)?

The problem is related, but this actually doesn't solve that. This solves a much worse bug in which the editor would get into an inconsistent state after that happened. Thanks for pointing out that issue though.

PR #211, opened 2017-02-16
Handle out-of-range folds when populating the spatial index

nathansobo:
When initially populating the spatial index, the row range isn't expanded to account for overlapping fold row ranges because the index isn't populated yet.

When `updateSpatialIndex` is called with an `oldEndBufferRow` exceeding the indexed buffer row count, we need to perform additional queries to the `foldsMarkerLayer` if folds exceed the requested buffer row range to ensure folds with intersecting row ranges are correctly inserted into the index.

This PR also adds a randomized test focusing exclusively on the issue of a pre-populated `foldsMarkerLayer`.

/cc @maxbrunsfeld


maxbrunsfeld:
👏  Nice!

PR #216, opened 2017-03-02
Emit marker layer update events synchronously at end of next transaction

nathansobo:
Previously, marker layer update events were emitted asynchronously in a `nextTick` callback to coalesce multiple marker updates into a single event. Unfortunately, this creates challenges writing rendering code for the editor because async animation frame requests interact in unpredictable ways with the next tick and make tests brittle. Async layer events also create a redundant update immediately following editor creation due to the cursors getting added. Keeping things sync simplifies code that subscribes to these events, while waiting until the end of the next transaction gives us tools to prevent these events from firing too frequently.

/cc @maxbrunsfeld @as-cii

### DIFF
--- a/spec/marker-layer-spec.coffee
+++ b/spec/marker-layer-spec.coffee
@@ -176,24 +176,34 @@ describe "MarkerLayer", ->
       expect(layer2.findMarkers(containsPoint: [0, 4])).toEqual [layer2Marker]
 
   describe "::onDidUpdate", ->
-    it "notifies observers asynchronously when markers are created, updated, or destroyed", (done) ->
+    it "notifies observers at the end of the outermost transaction when markers are created, updated, or destroyed", ->
+      [marker1, marker2] = []
+
       updateCount = 0
       layer1.onDidUpdate ->
         updateCount++
         if updateCount is 1
-          marker1.setRange([[1, 2], [3, 4]])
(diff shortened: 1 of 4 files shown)


spec/marker-layer-spec.coffee
```diff
@@ -176,24 +176,34 @@ describe "MarkerLayer", ->
...
 
   describe "::onDidUpdate", ->
-    it "notifies observers asynchronously when markers are created, updated, or destroyed", (done) ->
+    it "notifies observers at the end of the outermost transaction when markers are created, updated, or destroyed", ->
+      [marker1, marker2] = []
+
       updateCount = 0
       layer1.onDidUpdate ->
         updateCount++
         if updateCount is 1
```
as-cii: Super minor comment, but now that the event fires synchronously, do you think we could restructure this test to be simpler? Like:

```js
let updateCount = 0
layer1.onDidUpdate(() => updateCount++)

// first update
// second update
// etc.
```

nathansobo: I specifically want to test reentrant events... making marker updates from within events that fire additional events shouldn't cause infinite loops or stack overflows and should emit additional events.

as-cii: Oops, sorry the diff was a little bit deceiving, but it makes total sense now. :shipit:

PR #217, opened 2017-03-02
Using standard to linting javascript

torn4dom4n:
I think using `standard` better than `eslint`!


torn4dom4n:
@50Wliu Can you close this? I can't active it anymore!

PR #218, opened 2017-03-03
Always update display layer before emitting a marker layer update event

as-cii:
Previously, when a change was happening outside of a transaction (e.g. undo, redo, etc.), we were emitting a marker layer update event as soon as we spliced each marker layer. This was causing display layers to be out of date when the marker layer event occurred and could therefore cause problems if users performed a display layer query display layer inside the update event.

With this commit, marker layer update events will now always be emitted after updating display layers, thus fixing the errors we were observing in Atom (https://circleci.com/gh/atom/atom/2806).

/cc: @nathansobo


nathansobo:
Thanks so much for your help with this unexpected issue.

PR #220, opened 2017-03-06
Add encoding auto-detection logic

winstliu:
This is one out of four parts necessary to move encoding auto-detection logic into core, the other three being https://github.com/atom/node-pathwatcher/pull/120, https://github.com/atom/atom/pull/13943, and https://github.com/atom/encoding-selector/pull/42.

Specifically, this PR calls PathWatcher file APIs to detect the encoding.

This PR should be the second one to be merged.


winstliu:
https://github.com/atom/atom/pull/13943#issuecomment-930622479

PR #222, opened 2017-03-08
make one test fail to check Travis status

dirk-thomas:
**Do not merge this**

Just to check the Travis status when a spec is failing. See https://github.com/atom/text-buffer/pull/195#issuecomment-284924359


dirk-thomas:
I just rebased this against the latest master with the fix: d766103166b0257ed4cb19fddcf3e2e74a5e32d0

dirk-thomas:
Great, not green anymore as expected. Thanks for the quick fix.

Closing this since it was only demonstrating the problem...

maxbrunsfeld:
Thanks!

PR #223, opened 2017-03-14
Add oldText field to onDidChangeText events

hansonw:
This is useful for interfacing with LSP servers, as we need to know the length of text that was changed. Currently this is not easily obtainable from just the extents.

I also documented `onDidChangeText` as a public API.

Side note: these extents are in a weird format. Is this something broadly used within Atom? Does it make more sense to additionally return `oldRange` / `newRange` to be consistent with `onDidChange` and drop `oldExtent`/`newExtent` from the public documentation?


nathansobo:
@hansonw I've addressed your concern about the extents being a confusing interface by using `oldRange` and `newRange` instead. The old fields remain as undocumented, non-enumerable properties to prevent disruption of existing users.

I've also added more documentation describing these aggregated change events in the header of the class as well as all in the description of all methods that involve them. Want to check my work here and see if it's clear?

hansonw:
Woohoo! This looks much better, thank you so much for this.

maxbrunsfeld:
💸  Great docs @nathansobo.

nathansobo:
Testing out a build with this as a prerelease to minimize chaos.

nathansobo:
Just published as 11.4.0 and upgraded on`atom/atom#master`

src/text-buffer.coffee
```diff
@@ -1077,11 +1153,14 @@ class TextBuffer
...
-  # * `newExtent`: A {Point} representing the replacement extent.
-  # * `newText`: A {String} representing the replacement text.
+  # Returns an {Array} of {Object}s with the following fields that summarize
+  #  the aggregated changes since the checkpoint. See *Working With Aggregated
+  # Changes* in the description of the {TextBuffer} class for details.
+  # * `oldRange` The {Range} of the deleted text in the text as it existed when
+  #   the checkpoint was created.
+  # * `newRange`: The {Range} of the inserted text in the current text.
+  # * `oldText`: A {String} representing the deleted text.
+  # * `newText`: A {String} representing the ineserted text.
```
t9md: spell

s/ineserted/inserted/

Not only this one.

nathansobo:
Thanks for the heads up. Fixed on master so it will go out with the next published release.

PR #227, opened 2017-04-28
Clear cached screen lines correctly when range invalidation events occur

as-cii:
Previously, we were mistakenly translating the buffer range to screen coordinates and invalidating all the cached lines located at that screen range. This, however, could lead to emitting duplicate lines when folds or soft wrap occurred at the given invalidation range.

With this pull request we will find the line boundaries preceding and following the given buffer range, and splice the screen lines corresponding to those buffer rows instead.


as-cii:
It looks like we haven't received any bug report related to this issue, so we may cherry pick the fix to beta only, so that it goes to stable once we roll the railcars.

PR #232, opened 2017-05-10
Optimize critical code paths and add displayLayer.bufferRowsForScreenRows

as-cii:
This pull request addresses multiple concerns in the critical rendering code path:

1. Uses bitwise operations as opposed to modulo arithmetic to determine if numbers are odd or even.
2. Improves the performance of `compareBufferPosition` in ScreenLineBuilder by performing only one comparison and using subtraction instead.
3. Avoids de-optimized code paths due to "Unsupported phi use of const or let", replacing usages of `const` or `let` with `var`.
4. Adds a `DisplayLayer.prototype.bufferRowsForScreenRows` method that efficiently queries the spatial index to report every buffer row in the given screen row range.

/cc: @nathansobo

### DIFF
--- a/spec/display-layer-spec.js
+++ b/spec/display-layer-spec.js
@@ -2193,6 +2193,27 @@ describe('DisplayLayer', () => {
     })
   })
 
+  describe('.bufferRowsForScreenRows(startRow, endRow)', () => {
+    it('returns an array containing the buffer rows for the given screen row range', () => {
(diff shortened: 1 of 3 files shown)


spec/display-layer-spec.js
```diff
@@ -2193,6 +2193,27 @@ describe('DisplayLayer', () => {
     })
   })
 
+  describe('.bufferRowsForScreenRows(startRow, endRow)', () => {
```
nathansobo: ✨

nathansobo:
It's really too bad about the phi variable issue.

PR #233, opened 2017-05-11
Don't reset the "did-stop-changing" timeout for each buffer change

as-cii:
Instead, use a smarter debounce function that always waits for the existing timeout to fire and, if the buffer has changed in the meantime, schedules a new one.

/cc: @nathansobo


nathansobo:
Great idea.

PR #238, opened 2017-07-21
Make sure TextBuffer.save() calls through to custom files

hansonw:
The unit test didn't catch this since the path / file are otherwise identical 😆
I amended the test to actually ensure that `createWriteStream()` is being called.

Released under CC0.

### DIFF
--- a/src/text-buffer.coffee
+++ b/src/text-buffer.coffee
@@ -1552,7 +1552,7 @@ class TextBuffer
 
   saveTo: (file) ->
     if @destroyed then throw new Error("Can't save destroyed buffer")
-    unless file then throw new Error("Must provide a file to save")
+    unless file then throw new Error("Can't save a buffer with no file")
 
     filePath = file.getPath()
     if file instanceof File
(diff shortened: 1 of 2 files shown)


src/text-buffer.coffee
```diff
@@ -1539,7 +1539,10 @@ class TextBuffer
   #
   # Returns a {Promise} that resolves when the save has completed.
   save: ->
-    @saveAs(@getPath())
+    if @file?
+      @saveTo(@file)
+    else
+      @saveAs(@getPath())
```
maxbrunsfeld: I think this method could just do `@saveTo(@file)` with no conditional; it looks like if there is no `@file`, then `@getPath()` will return `null`, so `saveAs` will throw an exception anyway.

maxbrunsfeld:
Thanks for fixing this @hansonw!

src/text-buffer.coffee
hansonw: :+1:

hansonw:
The `"Can't save buffer with no file path"` message does seem a bit better if there's no file set. Maybe it's not too important, since you shouldn't be hitting this case normally?

maxbrunsfeld:
Oh, that's a good point. I'd like to consolidate logic into `saveTo` though. Maybe we should change the error message in that method to `Can't save a buffer with no file`.

PR #239, opened 2017-07-24
Support ES6 unicode regexp in scanInRange

lierdakil:
Patch itself is trivial. I had to add `js` implementation for spec though, since coffeescript doesn't support unicode regexp.


lierdakil:
No longer relevant, since TextBuffer apparently now uses pcre2 as a backend.

Issue #240, opened 2017-07-27
TextBuffer v13 fires spurious conflict & reload events


hansonw:
With the new `TextBuffer` implementation, a conflict event is always fired if a file change event is received with modified contents (source: https://github.com/atom/text-buffer/blob/master/src/text-buffer.coffee#L1750)

The prior (v12) implementation did something smarter, where it would check the new disk contents vs the previously saved contents and exit if it detected no real change. (https://github.com/atom/text-buffer/blob/v12.1.4/src/text-buffer.coffee#L1609)

Omitting the check is usually OK when the file change event is received before the save finishes, but this definitely isn't always the case (especially with the Nuclide remote filesystem use case). It's easy to reproduce this just by turning the debounce delay (`TextBuffer.fileChangeDelay`) way up, even with local files. I could definitely see a situation in which a user saves a file and types within the default 200ms delay, resulting in a spurious conflict.

hansonw:
@maxbrunsfeld what do you think would be the best path forward here? It sounds like bringing back `@cachedDiskContents` might not be a great change.. it sounds like we'd have to have some way of reading the disk contents without resetting the text buffer. Does the native API expose a way to do that?

Also, this results in spurious `did-reload` events as well, since every save event will result in a `did-reload` event as well with the new implementation. I'm not sure if that's intentional or not, but it's nice to be able to discern when a file actually had to be reloaded from disk (with changes). I'm fine with using `did-change` instead for this purpose, thoug.

PR #241, opened 2017-07-28
Avoid spurious conflicts when typing after saving

maxbrunsfeld:
Fixes https://github.com/atom/text-buffer/issues/240
Depends on https://github.com/atom/superstring/issues/19

/cc @hansonw This still needs some testing. I'm about to get on a plane; maybe we can talk through it on Monday.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "mkdirp": "^0.5.1",
     "pathwatcher": "7.1.0",
     "serializable": "^1.0.3",
-    "superstring": "^2.0.8",
+    "superstring": "^2.0.9-0",
     "underscore-plus": "^1.0.0"
   }
 }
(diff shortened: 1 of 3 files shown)


hansonw:
Awesome <3 It would be great if this could somehow avoid the extra `will-reload` / `did-reload` events as well, but I realize that it would be hard to implement that without resulting in a double-read. Anyway, I can live with switching to using the `did-change` event instead.

hansonw:
Actually, maybe that part can be implemented by only emitting `did-reload` if the contents actually changed (I dunno if that's desirable? Not sure what else the reload events are used for).

maxbrunsfeld:
Oh yeah, I can address the reload thing as well. We should not be emitting will-reload or did-reload events if the file is not changing.

PR #244, opened 2017-08-15
Use native text buffer to check if file is unmodified & to be deleted

kuychaco:
Addresses regression described in https://github.com/atom/atom/issues/14765

In the callback for the pathwatcher delete event we want to destroy the buffer if the underlying file has been deleted, unless the file has been modified. When checking for modification status, we need to call the [superstring](https://www.npmjs.com/package/superstring) native buffer's `isModified` method rather than the `isModified` method on this module's TextBuffer class which first checks if the file exists. This file existence check is unnecessary and unwanted since the file won't exist after a deletion.

/cc @maxbrunsfeld

### DIFF
--- a/spec/text-buffer-io-spec.js
+++ b/spec/text-buffer-io-spec.js
@@ -928,7 +928,6 @@ describe('TextBuffer IO', () => {
         })
 
         it('destroys the buffer', (done) => {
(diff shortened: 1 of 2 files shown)


BinaryMuse:
@maxbrunsfeld This looks good to me. Anything stand out to you?

PR #246, opened 2017-08-18
Restore the ability to save files owned by root on Mac

maxbrunsfeld:
Fixes https://github.com/atom/atom/issues/15267

The previous code for saving files owned by root relied on a synchronous native API: [`runas`](http://github.com/atom/node-runas). This PR adds a dependency on a new native module, [`spawn-as-admin`](https://github.com/atom/spawn-as-admin), which works very similarly to node's built-in `child_process.spawn` API, allowing us to save the file asynchronously.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2,129 +2,195 @@
   "name": "text-buffer",
   "version": "13.0.11",
   "lockfileVersion": 1,
+  "requires": true,
   "dependencies": {
     "@types/node": {
-      "version": "https://registry.npmjs.org/@types/node/-/node-7.0.41.tgz",
-      "integrity": "sha1-TbzprktVHfHvQEeQRgDFgU1Y4c4=",
+      "version": "7.0.42",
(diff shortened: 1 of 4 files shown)


as-cii:
> I'm not doing any of the sync calls here, because as far as I understand, the FileRecoveryService in Atom obviates the need for that backup logic. Is that right?

Yes, that should be the case. However, I remember people reporting missing recovery files after hard crashes, and the only scenario in which that would happen is:

1. Before saving the file, we copy the original version in the recovery folder.
2. We save the file to disk, without waiting for the OS to actually flush the bytes to disk (which is what I assume `sync` does).
3. We delete the recovery file.
4. A hard crash occurs.

So I think it might be interesting to experiment with `sync`, although it's hard to tell whether they fix the issue or not because reproducing the file loss behavior is quite tricky.

maxbrunsfeld:
Ok, I think I'm going to merge this as is, since it sounds like the `sync` concern is somewhat orthogonal to privilege escalation. We should talk more about how to employ `sync` though, since Atom 1.19 removed the `backUpBeforeSaving` option.

PR #249, opened 2017-08-29
Fix translation of screen row queries in `DisplayMarkerLayer.findMarkers`

as-cii:
Refs: https://github.com/atom/atom/issues/15441

Previously, when querying a display marker layer with the `startScreenRow`, `endScreenRow` or `intersectsScreenRowRange` parameters, we would simply convert the supplied screen rows into buffer rows and then query the underlying buffer marker layer using the `startRow`, `endRow` or `intersectsRowRange` parameters.

This behavior was, however, incorrect. In fact, translating a screen row into a buffer row and using that to perform the query in the buffer marker layer caused the entire buffer row to be scanned. In presence of folds or soft-wraps, such query could completely distort the area to scan that the caller was interested in, and possibly return markers outside of it or hiding markers inside of it.

With this pull-request we are changing the translation strategy so that:

* `startScreenRow` queries are translated using the new `startsInRange` query parameter in the buffer marker layer.
* `endScreenRow` queries are translated using the new `endsInRange` query parameter in the buffer marker layer.
* `intersectsScreenRowRange` queries are translated using the `intersectsRange` query parameter in the buffer marker layer.

### DIFF
--- a/src/marker-layer.coffee
+++ b/src/marker-layer.coffee
@@ -115,6 +115,12 @@ class MarkerLayer
           markerIds = filterSet(markerIds, @index.findStartingAt(Point.fromObject(value)))
         when 'endPosition'
           markerIds = filterSet(markerIds, @index.findEndingAt(Point.fromObject(value)))
+        when 'startsInRange'
+          {start, end} = Range.fromObject(value)
+          markerIds = filterSet(markerIds, @index.findStartingIn(start, end))
+        when 'endsInRange'
+          {start, end} = Range.fromObject(value)
+          markerIds = filterSet(markerIds, @index.findEndingIn(start, end))
         when 'containsPoint', 'containsPosition'
           position = Point.fromObject(value)
           markerIds = filterSet(markerIds, @index.findContaining(position, position))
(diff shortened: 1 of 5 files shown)


src/marker-layer.coffee
```diff
@@ -115,6 +115,12 @@ class MarkerLayer
           markerIds = filterSet(markerIds, @index.findStartingAt(Point.fromObject(value)))
         when 'endPosition'
           markerIds = filterSet(markerIds, @index.findEndingAt(Point.fromObject(value)))
+        when 'startsInRange'
```
as-cii: I purposefully avoided documenting `startsInRange` and `endsInRange` because they are only being used internally but I can definitely see exposing them at some point.

What do you think @maxbrunsfeld and @nathansobo? Should we wait a little more or should we go for it? 💭

nathansobo: I think you might as well document it. It's pretty low risk.

maxbrunsfeld:
If this needs to go into an Atom 1.19 hotfix, we'll need to cherry pick this PR to the `pathwatcher-7` branch of `text-buffer`. That branch is identical to master except that it uses an earlier version of `pathwatcher` so as to avoid a duplicate version of that module in Atom.

PR #254, opened 2017-09-12
Wait for promises returned by `onWillSave` handlers before saving

maxbrunsfeld:
Refs https://github.com/atom/event-kit/pull/36

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1930,9 +1930,9 @@
       "dev": true
     },
     "event-kit": {
-      "version": "2.3.0",
-      "resolved": "https://registry.npmjs.org/event-kit/-/event-kit-2.3.0.tgz",
-      "integrity": "sha1-RZugZG1Lfbyl2b8rPE4tAQPoXhU="
+      "version": "2.4.0",
+      "resolved": "https://registry.npmjs.org/event-kit/-/event-kit-2.4.0.tgz",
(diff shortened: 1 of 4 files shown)


szeck87:
Hi @maxbrunsfeld, does the promise simply need to be returned or does it have to be resolved?  I'm finding that returned rejected promises it's not saving the content of the buffer, though coffeescript isn't my favorite thing and could be a coding issue.  Also looking at the docs here:  https://atom.io/docs/api/v1.21.1/TextBuffer#instance-onWillSave it does say it will not save until the promise is resolved.

maxbrunsfeld:
We will now wait until promises returned from ‘onWillSave’ listeners to resolve before saving the file. If the listeners return a promise that *rejects*, then ‘buffer.save’ will reject with that error, and the file will not be saved.

szeck87:
OK thanks, one more question.  Is it possible to get the file path, or at least the file name and extension, after the save prompt for a file that hasn't been saved yet during onWillSave?  Basically:

1. Create new file
2. Add some content
3. Save it (save prompt appears)
4. Give the file a name/extension and where to save it

That all occurs before any onWillSave code is run, but I can't find any reference to what the user puts in there.

maxbrunsfeld:
Hmm, I don't think you can access that data currently. It would be possible to add it as an argument to the `onWillSave` listener though. Are you interested in submitting a PR to do that?

szeck87:
I could try, I've never looked at this code before nor actually built Atom locally.

PR #255, opened 2017-09-15
Allow custom history providers to be assigned

nathansobo:
This PR adds the following major methods:

* `setHistoryProvider` This method takes an object that can serve as the history implementation for the buffer.
* `getHistory` This returns data about the current state of the history so that a new history provider can be pre-populated and provide continuity from the previous provider.
* `restoreDefaultHistoryProvider` This takes a similar history snapshot to re-populate the history of the default provider.

@as-cii I realized we should probably add some documentation for these methods before merging this. I'll try to get that done later today.

### DIFF
--- a/spec/text-buffer-spec.coffee
+++ b/spec/text-buffer-spec.coffee
@@ -11,7 +11,7 @@ TextBuffer = require '../src/text-buffer'
 SampleText = fs.readFileSync(join(__dirname, 'fixtures', 'sample.js'), 'utf8')
 {buildRandomLines, getRandomBufferRange} = require './helpers/random'
 
-describe "TextBuffer", ->
+fdescribe "TextBuffer", ->
   buffer = null
 
   beforeEach ->
@@ -367,7 +367,7 @@ describe "TextBuffer", ->
       buffer.undo()
       expect(buffer.getText()).toBe "hello\nworld\r\nhow are you doing?"
 
-    it "does not allow the undo stack to grow without bound", ->
+    xit "does not allow the undo stack to grow without bound", ->
       buffer = new TextBuffer(maxUndoEntries: 12)
 
       # Each transaction is treated as a single undo entry. We can undo up
@@ -535,6 +535,7 @@ describe "TextBuffer", ->
         expect(buffer.getText()).toBe "hello\nworld\r\nhow are you doing?"
 
       it "groups adjacent transactions within each other's grouping intervals", ->
+        now += 1000
         buffer.transact 101, -> buffer.setTextInRange([[0, 2], [0, 5]], "y")
 
         now += 100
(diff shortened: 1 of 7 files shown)


paradoxxxzero:
Hi, I have rewritten my [undo-tree](/paradoxxxzero/atom-undo-tree) package to comply to this new history provider api.
If you have some time to spare, could you please take a look at [it](https://github.com/paradoxxxzero/atom-undo-tree/blob/master/lib/undo-tree.js)?
Is it how this new API is supposed to work?
Am I missing something on how to import `text-buffer` objects in a package? (Getting objects constructors seems a bit hackish)
Thanks in advance

nathansobo:
Hey @paradoxxxzero. Cool to see you taking interest in this API. I wasn't sure how useful it would be to the general public but I'm pleasantly surprised to see you take it up.

I took a quick look at your implementation. We weren't really intending the default provider to be subclassed, so you'll probably be a lot safer if you implement your history provider completely on your own and just conform to the existing interface. You could always base it on the current default implementation by copying code. That may seem like unfortunate duplication, but in the long run it will insulate you from any future churn in a class that isn't really intended to be subclassed and also give you more control.

If you have follow-up questions, you can ask them here.

paradoxxxzero:
Thank you very much for your review, as for code duplication I understand your point but I disagree with it. In this case the default implementation is the reference implementation for the interface you provided, this should not be encapsulated.

Consider I copy the file I want, I need nearly every other text-buffer files to be copied as well since the default implementation relies on them and a dozen of external dependencies. All that to change a little bit of the history implementation. Then, as I duplicated the entire project, I will have to merge all the bug fixes and features that you guys will make in this package. That would be an awful waste of time for a project of mine that won't change a lot (except when the api is broken).

I understand that encapsulation makes the developer life easier, but this have great impacts on atom extensibility, which is sad for the "hackable to the core" editor.

Thanks again for your time

nathansobo:
Yeah, I hear you. For now we're not going to go farther to make guarantees about subclassing the existing history provider as part of the API. But in practice your current approach of stealing the constructor should work for a while until we need to work on the history provider again. You might consider using composition rather than inheritance, where you wrap the default provider and delegate through to it as needed via methods in its public interface. Whatever works though. Just keep in mind it may break at some point in the future. With infinite time every dimension of Atom would be infinitely hackable, but we have to set limits while we devote attention to other things. You make good points though, so eventually we should come up with an officially supported solution for your use case.

paradoxxxzero:
I understand that it takes time to write and document apis and this PR is a good step forward, my undo-tree is now much more cleaner than before (it was a dirty dirty monkey patch: https://github.com/paradoxxxzero/atom-undo-tree/blob/ae15463edc3f7bb96a2efc8495f663ea2c85e050/lib/undo-tree.js).

Keep up the good work!

PR #257, opened 2017-09-18
Fix fromObject when `object` is undefined

bstream:
In atom@1.20.0 there is an unrecoverable error occurring when parameter `object` is undefined.

```
Uncaught (in promise) TypeError: Cannot read property 'start' of undefined
    at Function.module.exports.Range.fromObject (/Applications/Atom.app/Contents/Resources/app/node_modules/text-buffer/lib/range.js:33:37)
    at Function.module.exports.MarkerLayer.deserializeSnapshot (/Applications/Atom.app/Contents/Resources/app/node_modules/text-buffer/lib/marker-layer.js:39:57)
    at History.module.exports.History.deserializeStack (/Applications/Atom.app/Contents/Resources/app/node_modules/text-buffer/lib/history.js:427:60)
    at History.module.exports.History.deserialize (/Applications/Atom.app/Contents/Resources/app/node_modules/text-buffer/lib/history.js:361:35)
    at /Applications/Atom.app/Contents/Resources/app/node_modules/text-buffer/lib/text-buffer.js:264:30
```

NOTE: 
* This is _only_ happening in one of my projects and I do not know the root cause. This PR is just to protect against this error. 
* I could not try this fix locally since atom seems to cache its dependencies (inline edits did not do anything).


bstream:
If anyone else runs into this: This seems to happen due to a bad window state. Running atom with `--clear-window-state` when opening the project works fine for me.

PR #260, opened 2017-09-19
Introduce LanguageMode as a replacement for DisplayLayers' text decoration layers

maxbrunsfeld:
Currently, Atom's syntax highlighting system integrates with `TextBuffer` via a set of APIs based on the concept of a 'text decoration layer', which is described in the [original DisplayLayer PR](https://github.com/atom/text-buffer/pull/149). This PR restructures these APIs in the following ways:

#### The new name

The `TextDecorationLayer` interface has been renamed to `LanguageMode`. This is for two reasons:

1. The term 'text decorations' is overly general for what these objects are used for. We originally envisioned using these objects for more than just syntax highlighting, and allowing multiple of them to be added to a display layer. This capability was actually [implemented in `text-buffer`](https://github.com/atom/text-buffer/pull/228), but then removed, because we found that it actually *is* important to distinguish syntax highlighting from other forms of decorations. The more-general concept of text decorations has now been [implemented in Atom](https://github.com/atom/atom/pull/13880) in a different way.

2. In the effort to make Atom's language integration features more flexible, we want to move other language-related functionality like folding and auto-indent into a single object which also provides syntax highlighting. The `LanguageMode` will eventually be this object.

#### A property of the buffer, not the display layer

Currently, each `DisplayLayer` can have its own text decoration layer. This allows multiple text editor instances to simultaneously display a given file with different syntax highlighting. We believe this flexibility is unnecessary and possibly confusing. It also comes at a performance cost. We will now maintain a single language mode associated with a given buffer.

This PR doesn't change any public, documented APIs; only private one which are used by Atom.

/cc @nathansobo @as-cii


alexandernst:
Ooohhh tree-sitter is coming, oohhhh !!!!! \o/

PR #261, opened 2017-09-20
Emit text and marker change events when skipping undo

as-cii:
Back in #255 we introduced a regression that prevented the following events from firing when calling `setTextInRange` with `undo: 'skip'`: 

* `buffer.onDidChangeText`
* `buffer.onDidStopChanging`
* `markerLayer.onDidUpdate`
* `marker.onDidChange`

This pull-request fixes it by calling `emitMarkerChangeEvents` and `emitDidChangeTextEvent` at the end of `setTextInRange` if the transact call depth is `0` and `undo` is `'skip'`.

/cc: @nathansobo

### DIFF
--- a/spec/text-buffer-spec.coffee
+++ b/spec/text-buffer-spec.coffee
@@ -192,6 +192,70 @@ describe "TextBuffer", ->
         expect(buffer.undo()).toBe true
         expect(buffer.lineForRow(0)).toBe "hellow"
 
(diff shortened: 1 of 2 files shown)


nathansobo:
I see, because we don't `transact` if we're skipping undo and so the events that typically are fired at the end of the transaction are skipped. 👍

PR #263, opened 2017-09-26
Make isModified return true when buffer's file doesn't exist

maxbrunsfeld:
/cc @smashwilson

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "text-buffer",
-  "version": "13.3.4",
+  "version": "13.3.5-0",
   "lockfileVersion": 1,
   "requires": true,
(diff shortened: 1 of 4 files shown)


smashwilson:
Looks like the linter is unhappy:

```
/home/travis/build/atom/text-buffer/spec/text-buffer-io-spec.js
  580:7   error  'filePath' is not defined                            no-undef
  581:23  error  'filePath' is not defined                            no-undef
  581:38  error  Expected parentheses around arrow function argument  arrow-parens
```

smashwilson:
Okay, atom/autosave#84 is merged and shipped in autosave@0.24.6. I didn't bump it in atom/atom because these changes should likely go in together. :sparkles:

PR #265, opened 2017-10-05
Fix buffer coordinate translation for soft-wrapped atomic soft tabs

as-cii:
Previously, when translating buffer coordinates to screen coordinates we would first use the spatial index to retrieve the corresponding screen position and then apply a delta to it in order to clip the screen column to the left or to the right of an atomic token (such as a paired unicode character or an atomic soft tab).

The problem with this approach is that applying the delta to the screen column may result in a negative screen coordinate. In particular, this could happen if the translated buffer point is in the middle of an atomic soft-tab which has been soft-wrapped exactly at the given point. Consider the example below, assuming `.`s correspond to whitespaces:

```
Buffer Text:
....

Screen Text (Tab Length: 2, Soft Wrap Column: 3):
...
.
```

In the example above, translating `(0, 3)` to screen coordinates (assuming we clip backwards) should return `(0, 2)` but, since the clip delta is expressed in buffer coordinates, we used to return `(1, -1)`.

With this pull-request we are fixing the issue by clipping the buffer position *before* translating it to screen coordinates, thus avoiding the coordinate mismatch problem described above.

An alternative solution to this could be to prevent atomic soft tabs from being soft-wrapped, exactly as we do for paired unicode characters. I was torn on going down this path because it would be a less additive change, and would also result in a change in behavior. That said, it does feel more correct so I am open to revisiting this decision.

### DIFF
--- a/spec/display-layer-spec.js
+++ b/spec/display-layer-spec.js
@@ -883,46 +883,60 @@ describe('DisplayLayer', () => {
     })
 
     it('translates points correctly on soft-wrapped lines', () => {
-      const buffer = new TextBuffer({
-        text: '   abc defgh'
-      })
+      {
+        const buffer = new TextBuffer({text: '   abc defgh'})
+        const displayLayer = buffer.addDisplayLayer({
+          softWrapColumn: 8,
+          softWrapHangingIndent: 2
+        })
 
-      const displayLayer = buffer.addDisplayLayer({
-        softWrapColumn: 8,
-        softWrapHangingIndent: 2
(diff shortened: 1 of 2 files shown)


nathansobo:
Nice. Since we clip based on buffer positions it also makes more intuitive sense to apply this clipping first anyway. I think soft-wrapped atomic tabs is such a rare situation that it's not really worth spending additional time thinking about it right now.

nathansobo:
Published as 13.5.3.

PR #268, opened 2017-10-18
Document meaning of `Range` start and end points

AndiDog:
I was stuck trying to figure out what the start/end points actually mean while developing my first package which works with text selections. It's not immediately obvious that offsets are 0-based (yohoo!) and in which cases newlines are part of the range or not. Therefore I added a description and example tests to the documentation.


Arcanemagus:
> inclusive start and exclusive end

That part threw me off when I was first starting working with Atom, I'm glad somebody is documenting this!

50Wliu:
I'm :+1: on the better documentation, but I think adding a spec is a bit too much.  How about a text-based example instead?  Something like:
```
Given `01234`, the range `[[0, 1], [0, 3]]` will include `12`.  Newlines are not included unless the `end` {Point} is also on a new line.
```

AndiDog:
Made it text-based, just a little more verbose.

PR #277, opened 2017-11-01
Update setPath to work with custom files

hansonw:
Currently, `setPath` creates a new `File` object. If we already have a custom file object we should just call `setPath` on it, along with the other setup to update the subscriptions and send the event.

Contributed under CC0.


hansonw:
@maxbrunsfeld could you take a look at this when you get a chance? much appreciated :)

maxbrunsfeld:
Just so I understand - why is this preferable to just calling `setFile` with a new instance of the same file class? Up until now, I had envisioned `setPath(filePath)` as being a convenient shorthand for
`setFile(new pathwatcher.File(filePath))`.

hansonw:
Just convenience I guess - for example, creating remote files in Nuclide requires a fair amount of dependencies. It's a bit roundabout to have to create a new file when `File` implementations are able to change their own paths without much of an issue. The current behavior also means that we've effectively had to "blacklist" `TextEditor.setPath` from the Nuclide codebase, which seems a little odd.

PR #279, opened 2017-11-27
Report batched changes to display layers and language modes

maxbrunsfeld:
⚠️  WIP ⚠️


maxbrunsfeld:
Closing this out because I don't think it's ultimately a valid approach. Even though we don't emit events within transactions, the code that is *performing* a series of edits within a transaction is free to query the language mode or the display layer in the middle of the transaction, so the state needs to stay consistent.

PR #276, opened 2017-10-31
Show invisible spaces everywhere

50Wliu:
Addresses part of atom/atom#12157, specifically
> All displayed invisibles should be visible anywhere on the line when showInvisibles is enabled atom/atom#2572

PR #280, opened 2017-12-11
Fix compile script on Windows

### LINKED ISSUE #219: text-buffer does not install on Windows
[This command](https://github.com/atom/text-buffer/blob/master/package.json#L10) expects the existence of `cp`.  `cp` does not exist using Window's `cmd.exe`.

```cmd
> coffee --no-header --output lib --compile src && cp src/*.js lib/

'cp' is not recognized as an internal or external command,
operable program or batch file.
```

(Edit: the slashes also need to be backslashes)

Arcanemagus:
`npm` uses a `CMD` based environment for running scripts on Windows, which has no emulation of the `cp` command. Bring in a development dependency on `cpy-cli` and use that in the `compile` script to allow errorless installation on Windows.

Alternative options:
* Some Grunt script since it is already a dependency?
* [`copyfiles`](https://www.npmjs.com/package/copyfiles) is more downloaded, but `cpy-cli` seems to actually duplicate the behavior of `cp`

Fixes #219.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -79,14 +79,12 @@
     "array-find-index": {
       "version": "1.0.2",
       "resolved": "https://registry.npmjs.org/array-find-index/-/array-find-index-1.0.2.tgz",
-      "integrity": "sha1-3wEKoSh+Fku9pvlyOwqWoexBh6E=",
-      "dev": true
+      "integrity": "sha1-3wEKoSh+Fku9pvlyOwqWoexBh6E="
     },
     "array-union": {
       "version": "1.0.2",
       "resolved": "https://registry.npmjs.org/array-union/-/array-union-1.0.2.tgz",
       "integrity": "sha1-mjRBDk9OPaI96jdb5b5w8kd47Dk=",
-      "dev": true,
       "requires": {
         "array-uniq": "1.0.3"
       }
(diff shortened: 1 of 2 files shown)


Arcanemagus:
WARNING: That looks like an excessive amount of deletions in the `package-lock.json` to me...

50Wliu:
> WARNING: That looks like an excessive amount of deletions in the package-lock.json to me...

Ha, sounds like the same thing as what happened to me in #276.  Are you using npm 5.6.0?  I think that might fix it.

Also, fixes #219.  This will make my life much easier whenever I have to work on text-buffer 😍.

Arcanemagus:
> Are you using npm 5.6.0? I think that might fix it.

I am, so it must not fix it... if it's actually a problem?
```
PS C:\foo\text-buffer> npm --version
5.6.0
PS C:\foo\text-buffer> node --version
v9.2.1
```

Arcanemagus:
Rebased on `master`, the `package-lock.json` diff is a _lot_ saner this time around. The majority of the changes to this file are just removing `"dev": true` from a bunch of packages, the additions for `cpy-cli`, and updating of `superstring` to 2.2.17.

smashwilson:
👍

PR #281, opened 2017-12-12
Documentation fixes

Arcanemagus:
A number of documentation fixes, including:
* Document the required arguments to many of the checkpoint related functions
* Mark `::abortTransaction` as Public. This is mentioned as used in `::transact` so it should be marked Public as well.
* Documents return values on `undo` and `redo`
* Move mentions of "0-indexed" to the description instead of the arguments. A majority of the uses were already like this so this is just standardizing them all in this manner.
* Mark several arguments as optional that aren't necessary or have default values

A few notes for reviewers:
* The comment on the return value of `createCheckpoint` called it a "checkpoint value", but it is an ID. I fixed this as a _mistake_, but if it was intentional the wording of the new argument documentation in the other functions needs to be fixed. I think this might have been intentional to allow more flexibility in the implementation of a `HistoryProvider`.
* `abortTransaction` seems awkward, and I'm not sure the wording I've used is the best.
* There are several places in the Markers section that _could_ cross-link to the `Marker` and `MarkerLayer` entries, but I left that out for now.

I've left this as many small commits for easier review, let me know if you would prefer them squashed or split out into separate PRs!

### DIFF
--- a/src/text-buffer.coffee
+++ b/src/text-buffer.coffee
@@ -904,8 +904,8 @@ class TextBuffer
 
   # Public: Delete the lines associated with the specified 0-indexed row range.
   #
-  # If the row range is out of bounds, it will be clipped. If the startRow is
-  # greater than the end row, they will be reordered.
+  # If the row range is out of bounds, it will be clipped. If the `startRow` is
+  # greater than the `endRow`, they will be reordered.
   #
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
Thanks!

PR #284, opened 2018-01-08
Preserve marker identity when resurrecting a marker after undo/redo

nathansobo:
MarkerLayers can be created with a maintainHistory option, which allows the state of the marker layer to be restored to the state it had at the start/end of a transaction when undoing/redoing.

During this restoration process, a marker can be "resurrected". This can happen when you undo the transaction in which a marker was created and then redo it. Or it can happen when you undo a transaction in which a marker was destroyed.

Currently, when restoring a snapshot containing a marker that no longer exists, we create a *new* marker that mostly matches the state of the marker in the snapshot. However, this new marker does not have the same *id* as the previous marker. Also, even though markers are tracked in the index via their identifiers, we return markers as first-class objects in our API. Currently, because we create a new marker when resurrecting, anyone holding a reference to the old marker still sees their reference as destroyed. They're hanging on to stale state.

This PR changes our marker layer snapshots to include a reference to the marker object along with the snapshot of the marker's state. When restoring a marker that doesn't exist, we recycle the original object so that holders of existing references see their marker come back to life. This will allow snippets that don't contain mirrored tab stops to work correctly after undoing and redoing the snippet expansion, and just leads to a more coherent API overall.

It's worth noting that when *serializing* marker snapshots, we obviously can't include the object reference. This PR has conditional logic when serializing to ensure we don't include the marker reference in that situation, and we preserve the original behavior when restoring marker layer snapshots that don't include references, which happens during deserialization.

This PR also updates the Electron dependency used for interactive test runner to match the current Atom.

/cc @ungb

### DIFF
--- a/src/marker-layer.coffee
+++ b/src/marker-layer.coffee
@@ -290,7 +290,14 @@ class MarkerLayer
       if marker = @markersById[id]
         marker.update(marker.getRange(), snapshot, true, true)
       else
-        newMarker = @createMarker(snapshot.range, snapshot, true)
+        {marker} = snapshot
+        if marker
+          @markersById[marker.id] = marker
+          {range} = snapshot
+          @index.insert(marker.id, range.start, range.end)
+          marker.update(marker.getRange(), snapshot, true, true)
+        else
+          newMarker = @createMarker(snapshot.range, snapshot, true)
 
     for id in existingMarkerIds
       if (marker = @markersById[id]) and (not snapshots[id]?)
@@ -301,7 +308,7 @@ class MarkerLayer
     ranges = @index.dump()
     for id in Object.keys(@markersById)
       marker = @markersById[id]
-      result[id] = marker.getSnapshot(Range.fromObject(ranges[id]), false)
+      result[id] = marker.getSnapshot(Range.fromObject(ranges[id]))
     result
 
   emitChangeEvents: (snapshot) ->
@@ -314,7 +321,9 @@ class MarkerLayer
     markersById = {}
     for id in Object.keys(@markersById)
       marker = @markersById[id]
-      markersById[id] = marker.getSnapshot(Range.fromObject(ranges[id]), false)
+      snapshot = marker.getSnapshot(Range.fromObject(ranges[id]), false)
+      markersById[id] = snapshot
+
     {@id, @maintainHistory, @persistent, markersById, version: SerializationVersion}
 
   deserialize: (state) ->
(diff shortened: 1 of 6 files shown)


src/marker.coffee
```diff
@@ -382,8 +382,10 @@ class Marker
     @layer.markerUpdated() if updated and not suppressMarkerLayerUpdateEvents
     updated
 
-  getSnapshot: (range) ->
-    Object.freeze({range, @properties, @reversed, @tailed, @valid, @invalidate, @exclusive})
+  getSnapshot: (range, includeMarker=true) ->
+    snapshot = {range, @properties, @reversed, @tailed, @valid, @invalidate, @exclusive}
+    snapshot.marker = this if includeMarker
```
maxbrunsfeld: In the case where we include the marker, do you think it's worth exploring the approach of ditching the plain object altogether and using the marker itself as the snapshot? Like, in `MarkerLayer.createSnapshot`, we would 
replace `marker.getSnapshot(...)` with just `marker`.

I'm not sure if it'd be worth it. The reason I suggest it is that for the most part, the snapshot is already just a copy of the marker. The one additional property is `range`. I'm not sure how we'd deal with that... we used to store a property called `rangeWhenDestroyed` on destruction so that you could still query a destroyed marker's range.

src/marker-layer.coffee
```diff
@@ -301,7 +308,7 @@ class MarkerLayer
     ranges = @index.dump()
     for id in Object.keys(@markersById)
       marker = @markersById[id]
-      result[id] = marker.getSnapshot(Range.fromObject(ranges[id]), false)
+      result[id] = marker.getSnapshot(Range.fromObject(ranges[id]))
```
maxbrunsfeld: Should we be passing our `@maintainHistory` flag as the second argument to `getSnapshot` here, so that we only include the marker if we need it?

src/marker.coffee
maxbrunsfeld: Yeah, nevermind. I don't really have any ideas for a better way to do this actually.

nathansobo: Yeah, it's tough because the marker is mutable. I don't feel like this adds much overhead to the current snapshot. Just one more pointer to a shared object.

src/marker-layer.coffee
nathansobo: Nice idea.

nathansobo: Actually, the only call site is [already conditional](https://github.com/atom/text-buffer/blob/fix-marker-resurrection/src/text-buffer.coffee#L1957) on `maintainHistory` being true.

nathansobo:
To be safe, I'm going to cut a pre-release and test out an Atom build with this before merging.

src/marker-layer.coffee
maxbrunsfeld: Oh yeah, of course.

PR #286, opened 2018-01-11
Provide old language mode from onDidChangeLanguageMode

daviwil:
This change adds a second parameter to callbacks that are registered with `onDidChangeLanguageMode`.  This second parameter contains the buffer's previous language mode before the new mode was set.

This PR is related to atom/atom#13829 which reports that when saving an untitled file, the `softWrapped` setting that they changed using the "Toggle Soft Wrap" command gets reset once the file is saved.  A corresponding PR will be sent to atom/atom shortly.

### DIFF
--- a/spec/text-buffer-spec.coffee
+++ b/spec/text-buffer-spec.coffee
@@ -2611,7 +2611,7 @@ describe "TextBuffer", ->
       expect(buffer.getLanguageMode() instanceof NullLanguageMode).toBe(true)
 
       events = []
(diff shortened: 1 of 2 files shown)


daviwil:
Awesome, thanks!  How do version bumps usually work to get the updated changes back into atom/atom? Do we generally publish patch/minor releases for single PRs or do you want to batch a little more before bumping?

maxbrunsfeld:
I’d say go ahead and merge this and then bump the patch version.

daviwil:
Will do!

PR #287, opened 2018-01-15
Add exceptional snapshot handling for selectionsMarkerLayer

t9md:
Aiming to fix atom/atom#16176

### Scenario Which I want to fix(or improve)

When undo/redo, editor does not scroll to textual changes if editor is different editor from editor on which original change was made.
This situation easily happens when we opened multiple text-editors which shares same buffer.

### Condition

- Open `editor1`, make **change-1**
- Open `editor2` by copying `editor1`, make **change-2**
- Open `editor3` by copying `editor2`, make **change-3**
- Now `editor1`, `editor2`, `editor3` shares same buffer instance and have distinct selectionsMarkerLayer which is used to keep distinct cursors

# Current behavior

- Undo on `editor1`, `editor1`'s cursor is not restored to **point-of-before-change-3**.
- Undo on `editor2`, `editor2`'s cursor is restored to **point-of-before-change-2**, since `editor2` is same editor where **change-2** was made.
- Undo on `editor3`, `editor3`'s cursor is not restored to **point-of-before-change-1**.
- Open `editor4` by copying `editor3` and then destroy `editor1`, `editor2`, `editor3`
- Undoing/Redoing on `editor4` only undo/redo textual changes, but cursor is not restored, and not scrolled to textual changes, because `selectionsMarkerLayer` is not shared across editors.

# New behavior

NOTE: Require changes' in atom's text-editor to use new machanism

- Undo on `editor1`, `editor1`'s cursor is restored to **point-of-before-change-3** by restoring marker snapshot taken on `editor3`'s selectionsMarkerLayer to `editor1`.
- Undo on `editor2`, `editor2`'s cursor is restored to **point-of-before-change-2**.
- Undo on `editor3`, `editor3`'s cursor is restored to **point-of-before-change-1**.
- Open `editor4` by copying `editor3` and then destroy `editor1`, `editor2`, `editor3`
- Undoing/Redoing on `editor4` does undo/redo textual changes, and cursor is restored to each snapshot-ed point when change was made, because when undo/redo `selectionsMarkerLayer` is restored from different `selectionsMarkerLayer` which shares same buffer.

## Summary of changes

- buffer instance keep `selectionsMarkerLayerIds` as `Set`, it must be registered explicitly
- `transact`/`createCheckpoint`/`groupChangesSinceCheckpoint` take `selectionsMarkerLayerId`
  - Which eventually passed to `createMarkerSnapshot()` and skips snapshoting other `selectionsMarkerLayer` except passed id's.
  - Purpose of this skipping is to not restore cursor(=selections) which is not involved in changes
  - This skipping assure that snapshot does not contains multiple selectionsMarkerLayer's snapshot.
- `undo`/`redo` accept `selectionsMarkerLayerId`
  - Which restore snapshot taken in different `selectionsMarkerLayer` onto passed `selectionsMarkerLayerId`.
  - It’s like redirect restore target from originally taken layer to different layer

### DIFF
--- a/spec/marker-layer-spec.coffee
+++ b/spec/marker-layer-spec.coffee
@@ -171,25 +171,29 @@ describe "MarkerLayer", ->
       buffer.undo()
 
   describe "when role is provided for the layer", ->
-    it "getRole() returns it's role", ->
-      expect(Object.keys(buffer.markerLayerIdsByRole).length).toBe(0)
+    it "getRole() returns it's role and keep track of ids of 'selenctions' role", ->
+      expect(buffer.selectionsMarkerLayerIds.size).toBe 0
 
-      layerRoleA1 = buffer.addMarkerLayer(role: "role-a")
-      layerRoleB1 = buffer.addMarkerLayer(role: "role-b")
-      layerNoRole = buffer.addMarkerLayer()
-      expect(layerRoleA1.getRole()).toBe("role-a")
-      expect(layerRoleB1.getRole()).toBe("role-b")
-      expect(layerNoRole.getRole()).toBe(undefined)
+      selectionsMarkerLayer1 = buffer.addMarkerLayer(role: "selections")
+      expect(selectionsMarkerLayer1.getRole()).toBe "selections"
 
-      expect(buffer.markerLayerIdsByRole["role-a"].has(layerRoleA1.id)).toBe(true)
-      expect(buffer.markerLayerIdsByRole["role-a"].size).toBe(1)
-      expect(buffer.markerLayerIdsByRole["role-b"].has(layerRoleB1.id)).toBe(true)
-      expect(buffer.markerLayerIdsByRole["role-b"].size).toBe(1)
+      expect(buffer.addMarkerLayer(role: "role-1").getRole()).toBe "role-1"
+      expect(buffer.addMarkerLayer().getRole()).toBe undefined
 
-      layerRoleA2 = buffer.addMarkerLayer(role: "role-a")
-      expect(buffer.markerLayerIdsByRole["role-a"].has(layerRoleA1.id)).toBe(true)
-      expect(buffer.markerLayerIdsByRole["role-a"].has(layerRoleA2.id)).toBe(true)
-      expect(buffer.markerLayerIdsByRole["role-a"].size).toBe(2)
+      expect(buffer.selectionsMarkerLayerIds.size).toBe 1
+      expect(buffer.selectionsMarkerLayerIds.has(selectionsMarkerLayer1.id)).toBe true
+
+      selectionsMarkerLayer2 = buffer.addMarkerLayer(role: "selections")
+      expect(selectionsMarkerLayer2.getRole()).toBe "selections"
+
+      expect(buffer.selectionsMarkerLayerIds.size).toBe 2
+      expect(buffer.selectionsMarkerLayerIds.has(selectionsMarkerLayer2.id)).toBe true
+
+      selectionsMarkerLayer1.destroy()
+      selectionsMarkerLayer2.destroy()
+      expect(buffer.selectionsMarkerLayerIds.size).toBe 2
+      expect(buffer.selectionsMarkerLayerIds.has(selectionsMarkerLayer1.id)).toBe true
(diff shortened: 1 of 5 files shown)


iolsen:
@maxbrunsfeld can you take a look at this one?

maxbrunsfeld:
I looked into what other text editors do about this:

* **Sublime Text** seems to behave like Atom does now. Each editor for a given file has its own history of cursor positions, and they do not influence each other during `undo` or `redo` commands.
* **Vim** behave the way that you describe, @t9md. If I perform an `undo` in *Editor 1* that undoes a change that was initially made in *Editor 2*, the cursor position in *Editor 1* will be updated to match the previous cursor position in *Editor 2*.
* **VS Code** seems to behave similarly to vim in this regard.

I agree with you that the latter behavior is probably more useful.

maxbrunsfeld:
Overall, I like your approach of associating a particular marker layer with a given call to `transact`, `undo`, and `redo`. I think this marker layer could be referred to as the 'active marker layer'.

I don't think we would want to use the name `selectionsMarkerLayer` in the `text-buffer` library, because 'selection' is a higher-level concept that is specific to the Atom application.

Maybe we could generalize the concept in the following way:

Marker layers have a new optional field called `role` whose value is an arbitrary string (e.g. `"atom.selection"`). During a transaction, if an `activeMarkerLayer` is passed to `transact()`, we will record a marker snapshot for the active marker layer, but *skip* snapshotting any other marker layers with the same `role` as the active marker layer. So in the case of Atom, we will only snapshot the selections in the active text editor, not the selections in other editors for the same buffer.

Then, when `undo()` or `redo()` is called, if an `activeMarkerLayer` is passed, we will restore that layer's marker based on any layer in the snapshot that has the same `role`.

So to summarize my proposed API changes:

* `TextBuffer.addMarkerLayer` takes a new option: `role`.
* `MarkerLayer` has a new method: `getRole()`.
* `TextBuffer.transact` takes a new option: `activeMarkerLayer`.
* `TextBuffer.undo` and `TextBuffer.redo` now take an options object with one option: `activeMarkerLayer`.

Thoughts? Are there edge cases that I'm missing?

t9md:
I agree overall idea and API change you suggest, thanks!

I too thought of this `role` idea while working on this PR but avoid early abstraction and choose purpose explicit code since I was still not sure how team think this change.

> @t9md Thoughts? Are there edge cases that I'm missing?

Currently the **role** we can think of is just `atom.selection`.
But possibly more than two roles involved in futures in single transaction.
Should we take multiple `activeMarkerLayer` as argument?
Say `transact(groupInterval, callback, [activeMarkerLayer(role=atom.selection), activeMarkerLayer(role=atom.some-role)])`.
Not sure, might be OK to postpone this until we find other roles.

t9md:
@maxbrunsfeld

I noticed we still need to keep `Set` of markerLayer ids by `@markerLayerIdsByRole[role]`.
When change originator editor was destroyed and associated marker layer was destroyed too.
History still have snapshot taken on that destroyed marker layer and should be able to restore.

Described behavior at [restore snapshot taken at destroyed marker layer](https://github.com/atom/text-buffer/pull/287/commits/a65796519b108d469adb33e89f560add2ef5da51).

nathansobo:
Overall, this change makes sense. We really want to honor the selections of the editor making the change when snapshotting markers.

I'm wondering if we actually should just be explicit about a "selections marker layer" in the text buffer API. It's designed to support an editor and we're the primary consumer of the library. The advantage to doing it this way is we can keep the code simple and avoid unexpected combinations. The API is also clearer... "for the purposes of this transaction, these are my selections."

t9md:
OK, will try to be more specific.
- Keep makerLayer "role" capability, but use internally to distinguish "atom.selections" role of markerLayer
- transact/undo/redo take selectionsMarkerLayerId to inform which selectionsMarkerLayer to snapshot/restore to.

t9md:
@maxbrunsfeld , @nathansobo 

Updated PR to make selective snapshotting/restoration for "selections" role specific.

ungb:
@nathansobo any chance you can take a 👀 at this again?

smashwilson:
👍 Thanks!

PR #288, opened 2018-01-17
Improve handling of buffers that are too big to convert to strings

maxbrunsfeld:
Fixes https://github.com/atom/atom/issues/15325

V8 Strings have a fixed maximum length. For Atom to work with arbitrarily large files, we need to avoid calling methods like `TextBuffer.getText()`, which returns a buffer's entire text as a string. This PR improves our handling of buffers that exceed this length threshold: 

  1. Previously, the `TextBuffer.scanInRange` family of methods still used `.getText()` when handling regexes that could span multiple lines. Now, these methods are implemented in terms of superstring's native regex search capabilities. (Depends on https://github.com/atom/superstring/pull/49).
  Notably, these are the APIs used by [find-and-replace](https://github.com/atom/find-and-replace). So this will improve performance of searching large files.

2. Previously, when `getText` was called on one of these large buffers, the error thrown had this message: `Couldn't convert text to a String`. Now there is a message specifically recommending using other APIs instead of `.getText` (https://github.com/atom/superstring/commit/fa2aa3767a77f1845891110790f93b9dbf689f14#diff-cda2ff0962a6b0a4f3514e9541b323afR263)

3. The `TextBuffer` now has a `.getLength` method, which returns the number of UTF16 code units in the buffer. Previously, there was no public way of determining this without calling `getText()`.


nathansobo:
Awesome work.

yoursweater:
I'm trying to edit a 600mb json file and I'm getting this error still.

maxbrunsfeld:
@yoursweater this was just merged and hasn’t been published in a new atom release yet. Did you build Atom from source? If so, could you share your Atom SHA you built from and the stack trace?

PR #289, opened 2018-01-17
Add TextBuffer.findAndMarkAllInRangeSync method

maxbrunsfeld:
Depends on https://github.com/atom/superstring/pull/50


src/text-buffer.coffee
```diff
@@ -1490,6 +1491,33 @@ class TextBuffer
...
+    @nextMarkerId += @buffer.findAndMarkAllSync(
+      markerLayer.index,
+      startId,
+      exclusive,
+      regex,
+      Range.fromObject(range)
+    )
+    markers = []
+    id = startId
+    while id < @nextMarkerId
```
nathansobo: Too bad we have to instantiate these.

PR #292, opened 2018-03-06
Add emitDidChangePath as helper method for teletype bug 147

odoyle71:
**Requirements**

Filling out the template is required. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion.
All new code requires tests to ensure against regressions

**Description of the Change**
In a teletype Issue, [147](https://github.com/atom/teletype/issues/147), the guest's tab names (paths) are not reflecting the host's names when the host changes their path (i.e., `Save as...`). To remedy this, when the path changes, we need to access the Guest's `Text-Buffer` and have the buffer emit a `path-did-change` message. Since there is no actual file on the guest's filesystem for this, the `setPath` function does not work.  @lee-dohm recommended that we add a helper function (so that _Teletype_ does not access the _TextBuffer_ emitter directly).
We are adding the helper function `emitDidChangePath` (no parameters) to call this emitter.
This PR is a precursor to two future PRs in teletype and teletype-client.

**Alternate Designs**
We considered accessing the _TextBuffer_'s emitter directly through `this.buffer.emitter.emit('did-change-path', null)`, but realized that this is dangerous to do, since if _Text-Buffer_'s code is refactored, then _Teletype_ may have issues.

**Why Should This Be In Core?**
Adding this functionality is a prerequisite for a solution to a bug in Teletype, which is a key package, to be fixed properly. In particular this incorporates regression tests to ensure that the solution will hold up to code refactoring.

**Benefits**
[147](https://github.com/atom/teletype/issues/147) will be able to be resolved.

**Possible Drawbacks**
We have not yet noticed any drawbacks to this change, as it is simply adding and a function and not modifying existing code.

**Verification Process**
We have locally tested that the new function properly emits the correct message to have all TextEditors update their titles.
We have added two tests to the `text-buffer-spec.coffee` file and ensured that they work, using `atom --test.
`
**Applicable Issues**
[147](https://github.com/atom/teletype/issues/147)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3864,9 +3864,9 @@
       }
     },
     "superstring": {
-      "version": "2.2.13",
-      "resolved": "https://registry.npmjs.org/superstring/-/superstring-2.2.13.tgz",
-      "integrity": "sha512-nizh7J3/wMzXAO1LLwad/GzyWK6ELf56UkLbmcCJmtd6aS0ml2lcPKbULPS5xZDp8p9Gk5nVGQKQUrftgGs5+A==",
+      "version": "2.2.17",
+      "resolved": "https://registry.npmjs.org/superstring/-/superstring-2.2.17.tgz",
+      "integrity": "sha512-7/3jvvetxkJvVpoJwtA+30aha/COezEppfjESOAJvXMDtCY2UkDdgB5B7ltCr0KnCqMhhXOdSsI+ahgfqLfJ0Q==",
       "requires": {
         "nan": "2.6.2"
       }
--- a/package-lock.json
+++ b/package-lock.json
@@ -3864,9 +3864,9 @@
       }
     },
     "superstring": {
-      "version": "2.2.17",
-      "resolved": "https://registry.npmjs.org/superstring/-/superstring-2.2.17.tgz",
-      "integrity": "sha512-7/3jvvetxkJvVpoJwtA+30aha/COezEppfjESOAJvXMDtCY2UkDdgB5B7ltCr0KnCqMhhXOdSsI+ahgfqLfJ0Q==",
+      "version": "2.2.13",
+      "resolved": "https://registry.npmjs.org/superstring/-/superstring-2.2.13.tgz",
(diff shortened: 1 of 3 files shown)


as-cii:
Hello @odoyle71 and thanks for this pull request!

Adding a method such as `emitDidChangePath` feels somewhat unintuitive and, even though I agree that calling `emitter.emit` directly would be fragile, I also think that adding a method that just wraps it would not improve things significantly, as it would not hide the complexity of path handling behind an abstraction.

Thus, may I suggest you to experiment with a different approach? I am wondering if we could take advantage of `TextBuffer`'s experimental `setFile` API in Teletype. This method is already being used for manipulating remote files [in Nuclide](https://nuclide.io/docs/features/remote) and I believe it could be useful for Teletype as well.

Specifically, I think we could replace [Teletype's monkey-patches to `TextBuffer`](https://github.com/atom/teletype/blob/b9268e520931db2a2c9c0b13812eacfe3ea760f2/lib/editor-binding.js#L75-L76) with a call to `buffer.setFile(file)` and supply an object that implements the following methods:

* `getPath() -> string`, which can be used to return the remote buffer's path.
* `createWriteStream() -> WriteStream`, which can be used to perform calls to `BufferProxy::requestSave`.
* `onDidRename(callback) -> void`, which can be used to invoke the supplied `callback` whenever the host notifies guests that a certain buffer's path has changed. Implementing this method would essentially allow you to implement the feature described in https://github.com/atom/teletype/issues/147.

What do you think?

/cc: @jasonrudolph

CaptainJohnyAppleSeed:
Pardon our tardy response, I agree that it'd be best to implement a "dummy" file that implements those functions. We'd misunderstood the documentation, and thought that there had to be a real `path` or `file` that had to be submitted for this function to work. We will update and resubmit, thanks!

as-cii:
Awesome, I am going to close this for now and we can re-open it or create a new pull request if needed.

Thanks for your help on this, @CaptainJohnyAppleSeed! ⚡️

PR #293, opened 2018-03-14
Doc change for setFile (existsSync is necessary)

CaptainJohnyAppleSeed:
existsSync method is necessary for the object passed into setFile. Should be reflected in comment atom/text-buffer/issues/258

### DIFF
--- a/src/text-buffer.coffee
+++ b/src/text-buffer.coffee
@@ -576,6 +576,7 @@ class TextBuffer
   #   * `getPath` A {Function} that returns the {String} path to the file.
   #   * `createReadStream` A {Function} that returns a `Readable` stream
   #     that can be used to load the file's content.
+  #   * `existsSync` A {function} that returns the {Boolean}, true if the file exists, false otherwise
   #   * `createWriteStream` A {Function} that returns a `Writable` stream
   #     that can be used to save content to the file.
   #   * `onDidChange` (optional) A {Function} that invokes its callback argument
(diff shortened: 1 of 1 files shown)


src/text-buffer.coffee
```diff
@@ -576,6 +576,7 @@ class TextBuffer
   #   * `getPath` A {Function} that returns the {String} path to the file.
   #   * `createReadStream` A {Function} that returns a `Readable` stream
   #     that can be used to load the file's content.
+  #   * `existsSync` A {function} that returns the {Boolean}, true if the file exists, false otherwise
```
Arcanemagus: Can this be moved below `createWriteStream` so it doesn't split the two `create___Stream` properties up?

CaptainJohnyAppleSeed: Great catch, thanks! I've reordered them, such that they write streams are together.

src/text-buffer.coffee
```diff
@@ -578,6 +578,7 @@ class TextBuffer
   #     that can be used to load the file's content.
   #   * `createWriteStream` A {Function} that returns a `Writable` stream
   #     that can be used to save content to the file.
+  #   * `existsSync` A {function} that returns the {Boolean}, true if the file exists, false otherwise
```
Arcanemagus: Whoops, caught these on a last check:
* `{function}` should be `{Function}`
* Add a `.` on the end
* `returns the {Boolean}` -> `returns a {Boolean}`

CaptainJohnyAppleSeed:
Great catches again, thanks!

PR #295, opened 2018-04-05
Clarify deprecation of `undo: skip` option

jasonrudolph:
https://github.com/atom/text-buffer/pull/282 updated the docs for `TextBuffer::setTextInRange` to deprecate the `undo` option. https://github.com/atom/atom/issues/16956 points out that there are other public methods whose docs reference the `undo` option. This pull request updates the docs for those other methods so that they also identify the `undo` option as deprecated. It also emits a deprecation warning whenever the `undo` option is used.

### DIFF
--- a/src/text-buffer.coffee
+++ b/src/text-buffer.coffee
@@ -843,7 +843,9 @@ class TextBuffer
   # * `text` A {String} representing the text to insert.
   # * `options` (optional) {Object}
   #   * `normalizeLineEndings` (optional) {Boolean} (default: true)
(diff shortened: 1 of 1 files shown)


jasonrudolph:
To verify that the deprecation warning is wired up correctly, I ran Atom in dev mode and executed the following code in the console:

```js
atom.workspace.getActiveTextEditor().getBuffer().setTextInRange([[0,0], [0,1]], 'foo', {undo: 'skip'})
```

Afterwards, Atom correctly reports the use of this deprecated option:

image

PR #296, opened 2018-04-09
Move cpy-cli to a devDependency

Arcanemagus:
The `cpy-cli` dependency is only used in the `compile` script (which is called by `prepublish`), as such it should have been a `devDependency` from the start but was mistakenly added as a regular `dependency` in #280. This moves it to where it should have been.

Fixes https://github.com/atom/atom/issues/17125.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -79,12 +79,14 @@
     "array-find-index": {
       "version": "1.0.2",
       "resolved": "https://registry.npmjs.org/array-find-index/-/array-find-index-1.0.2.tgz",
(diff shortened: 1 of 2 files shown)


Arcanemagus:
The failing test doesn't appear to be related to this change at all, since the code would either die before even starting, or work the exact same as before from this.

Arcanemagus:
Looks like [this build](https://travis-ci.org/atom/text-buffer/builds/364197885#L1396) is showing a flaky test, rebased on `master` and everything is fine.

PR #303, opened 2018-09-17
Preserve buffer markers on non-owned MarkerLayers

smashwilson:
When a DisplayMarkerLayer is destroyed, it unconditionally clears the markers from its MarkerLayer, whether or not that MarkerLayer is owned. This is surprising to me because the MarkerLayer itself persists; it feels like the markers on that layer should be preserved, as well.

```js
const {TextEditor, TextBuffer} = require('atom')

const buffer = new TextBuffer();
buffer.retain();
const layer = buffer.addMarkerLayer();
const marker = layer.markRange([[0, 0], [0, 0]]);

const editor = new TextEditor({buffer});
const displayMarkerLayer = editor.getMarkerLayer(layer.id);
editor.decorateMarkerLayer(layer, {type: 'line', class: 'whatever'});
editor.destroy();

console.log(displayMarkerLayer.isDestroyed()); // true, ok
console.log(layer.isDestroyed()); // false, ok
console.log(marker.isDestroyed()); // true?

buffer.release();
```

I noticed this in atom/github#1512 because I noticed that buffer markers on decorated MarkerLayers were being destroyed when a TextEditor that rendered them was closed. This is causing me problems because I'm caching and re-using TextBuffers and their markers.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,7 +22,7 @@
       "integrity": "sha1-r9+UiPsezvyDSPb7IvRk4ypYs2s=",
       "dev": true,
       "requires": {
-        "acorn": "3.3.0"
+        "acorn": "^3.0.4"
       },
       "dependencies": {
         "acorn": {
@@ -39,8 +39,8 @@
       "integrity": "sha1-gv+wKynmYq5TvcIK8VlHcGc5xTY=",
       "dev": true,
       "requires": {
-        "co": "4.6.0",
-        "json-stable-stringify": "1.0.1"
+        "co": "^4.6.0",
+        "json-stable-stringify": "^1.0.1"
       }
     },
     "ajv-keywords": {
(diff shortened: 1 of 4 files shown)


smashwilson:
@maxbrunsfeld Is this a bug, or is this how DisplayMarkerLayers and MarkerLayers are supposed to work?

maxbrunsfeld:
It does seem like a bug, though it might have some unexpected implications. 

Maybe we should publish a prerelease off this branch and open an Atom PR to upgrade it, just to see if Atoms full CI catches any unexpected fallout.

smashwilson:
> Maybe we should publish a prerelease off this branch and open an Atom PR to upgrade it, just to see if Atoms full CI catches any unexpected fallout.

:+1:

Published as text-buffer@13.14.9-0 and testing in atom/atom#18067. Let's see if that's green 🤞

smashwilson:
Also: I can confirm with a dev build that this fixes the issue I was seeing in atom/github#1512.

smashwilson:
Upstream VSTS builds were green 👍

PR #311, opened 2019-05-28
Escalate privileges when writing to an unauthorized file also on Linux

as-cii:
Refs https://github.com/atom/atom/issues/4115 and https://github.com/atom/atom/pull/19412

- [x] Depends on https://github.com/atom/fs-admin/pull/15

As the title says, this pull request will enable privilege escalation on Linux when attempting to save a file that the current user does not have access to.


as-cii:
I tested this on Ubuntu and Fedora, and it seems to work! Merging and publishing a new version so that we can use it in Atom.

PR #318, opened 2020-03-16
Fix project-level find-and-replace regex bug

darangi:
### Description of the Change
This PR ensures regex literals are properly replaced during project-level find-and-replace with the regex option selected e.g `atom/flight-manual.atom.io`  is replaced with `atom/flight-manualatomio` instead of `atom/flight-manual1$io`  when the search string is `/\.(atom)\./` and the replacement string is `1$`
### Applicable Issues

Fixes the issue in find-and-replace package  https://github.com/atom/find-and-replace/issues/1110

### DIFF
--- a/spec/text-buffer-spec.js
+++ b/spec/text-buffer-spec.js
@@ -24,7 +24,7 @@ describe('when a buffer is already open', () => {
 
     it('replaces atom/flight-manualatomio with $1', () => {
       buffer.setText('atom/flight-manualatomio')
-      buffer.replace($1, '$1')
+      buffer.replace(/\.(atom)\./, '$1')
 
       expect(buffer.getText()).toBe('atom/flight-manualatomio')
     })
--- a/spec/text-buffer-spec.js
+++ b/spec/text-buffer-spec.js
@@ -2,9 +2,10 @@ const path = require('path')
 const TextBuffer = require('../src/text-buffer')
 
 describe('when a buffer is already open', () => {
+  const filePath = path.join(__dirname, 'fixtures', 'sample.js')
+  const buffer = new TextBuffer()
+
   it('replaces foo( with bar( using /\bfoo\\(\b/gim', () => {
-    const filePath = path.join(__dirname, 'fixtures', 'sample.js')
-    const buffer = new TextBuffer()
     buffer.setPath(filePath)
     buffer.setText('foo(x)')
     buffer.replace(/\bfoo\(\b/gim, 'bar(')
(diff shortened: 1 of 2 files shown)


spec/text-buffer-spec.js
```diff
@@ -2,13 +2,31 @@ const path = require('path')
...
-    const buffer = new TextBuffer()
     buffer.setPath(filePath)
     buffer.setText('foo(x)')
     buffer.replace(/\bfoo\(\b/gim, 'bar(')
 
     expect(buffer.getText()).toBe('bar(x)')
   })
+
+  describe('should properly replace regex literals', () => {
+    it('replaces tstat_fvars()->curr_setpoint[HEAT_EN] = new_tptr->heat_limit; with tstat_set_curr_setpoint(HEAT_EN, new_tptr->heat_limit);', () => {
```
lee-dohm: It's not easy to understand what this test case is supposed to do. Can we simplify it?

spec/text-buffer-spec.js
```diff
@@ -2,13 +2,31 @@ const path = require('path')
...
+      buffer.setPath(filePath)
+      buffer.setText('tstat_fvars()->curr_setpoint[HEAT_EN] = new_tptr->heat_limit;')
+      buffer.replace(/tstat_fvars\(\)->curr_setpoint\[(.+?)\] = (.+?);/, 'tstat_set_curr_setpoint($1, $2);')
+
+      expect(buffer.getText()).toBe('tstat_set_curr_setpoint(HEAT_EN, new_tptr->heat_limit);')
+    })
+
+    it('replaces atom/flight-manual.atom.io with atom/flight-manualatomio', () => {
+      buffer.setText('atom/flight-manual.atom.io')
```
lee-dohm: Does the project-level find-and-replace open a `TextBuffer` object for every file that is operated on?

The bug appears to be in the project-level replacements and I believe this `TextBuffer` functionality is only utilized on the file-level replacements? Or am I misunderstanding?

darangi: `TextBuffer` is used on project level. The bug exists whenever you try to replace the contents in an open buffer. if the file is not opened, the replacement works fine.

lkashef:
Fixes https://github.com/atom/find-and-replace/issues/1110

eliezertor:
This bug is still happening, hopefully an update will be coming soon to fix this issue.

Aerijo:
@eliezertor If you are referring to it still happening in Atom stable, then it will be a while before the change propagates down to there.

If you want to receive bug fixes and features much sooner, try using the [beta channel](https://atom.io/beta) (relatively stable, releases timed with stable) or the [nightly channel](https://atom.io/nightly) (least stable, but get changes within a few days).

The more users using these pre release channels, the sooner bugs like these can be caught and fixed.

PR #326, opened 2020-09-30
test on electron 6

aminya:
### Description of the change

Updates electron devDependency which is used to run the tests. This version is the same as the electron used in Atom.

### Verification
This package is already used on Electron 6 inside Atom.

Closes #323


aminya:
@sadick254 Check this out

sadick254 (approved):

PR #331, opened 2020-10-15
Allow to save hidden files in Windows

gosukiwi:
### Identify the Bug

This PR addresses https://github.com/atom/atom/issues/13538

### Description of the Change

Trying to save a hidden file in Windows will raise a permission error. This PR uses [winattr](https://www.npmjs.com/package/winattr) to bypass that issue by toggling the hidden attribute to false while saving, and then turning it back on when done.

### Alternate Designs

I do introduce a dependency (winattr) as there's no easy way to handle attributes for files in Windows using Node.js. Alternatively, it could be written by hand, but it's a lot of tests and code that is better delegated to it's own library. Winattr is quite popular and stable so I think it's by far the best option.

### Possible Drawbacks

--

### Verification Process

I tried it locally on my Windows 10 installation and it worked. I also added tests and they passed.

### Release Notes

TextBuffer is now able to save hidden files in Windows

### DIFF
--- a/src/text-buffer.js
+++ b/src/text-buffer.js
@@ -1927,13 +1927,28 @@ class TextBuffer {
       try {
         await this.buffer.save(destination, this.getEncoding())
       } catch (error) {
-        const canEscalate = process.platform === 'darwin' || process.platform === 'linux'
-        if (error.code === 'EACCES' && destination === filePath && canEscalate) {
-          const fsAdmin = require('fs-admin')
-          try {
-            await this.buffer.save(fsAdmin.createWriteStream(filePath), this.getEncoding())
-          } catch (_) {
-            throw error
+        const isMacOrLinux = process.platform === 'darwin' || process.platform === 'linux'
+        const isWindows = process.platform === 'win32'
+        if (error.code === 'EACCES' && destination === filePath) {
+          if (isMacOrLinux) {
+            const fsAdmin = require('fs-admin')
+            try {
+              await this.buffer.save(fsAdmin.createWriteStream(filePath), this.getEncoding())
+            } catch (_) {
+              throw error
+            }
+          } else if (isWindows) {
+            const winattr = require('winattr')
+            const attrs = winattr.getSync(filePath)
+            if (!attrs.hidden) throw error
+
+            try {
+              winattr.setSync(filePath, { hidden: false })
+              await this.buffer.save(filePath, this.getEncoding())
+              winattr.setSync(filePath, { hidden: true })
+            } catch (_) {
+              throw error
+            }
           }
         } else {
           throw error
--- a/src/text-buffer.js
+++ b/src/text-buffer.js
@@ -1929,16 +1929,8 @@ class TextBuffer {
       } catch (error) {
         if (error.code !== 'EACCES' || destination !== filePath) throw error
 
-        const isMacOrLinux = process.platform === 'darwin' || process.platform === 'linux'
         const isWindows = process.platform === 'win32'
-        if (isMacOrLinux) {
-          const fsAdmin = require('fs-admin')
-          try {
-            await this.buffer.save(fsAdmin.createWriteStream(filePath), this.getEncoding())
-          } catch (_) {
-            throw error
-          }
-        } else if (isWindows) {
+        if (isWindows) {
           const winattr = getPromisifiedWinattr()
           const attrs = await winattr.get(filePath)
           if (!attrs.hidden) throw error
@@ -1947,6 +1939,13 @@ class TextBuffer {
             await winattr.set(filePath, { hidden: false })
             await this.buffer.save(filePath, this.getEncoding())
             await winattr.set(filePath, { hidden: true })
+          } catch (_) {
+            throw error
+          }
+        } else {
+          const fsAdmin = require('fs-admin')
+          try {
+            await this.buffer.save(fsAdmin.createWriteStream(filePath), this.getEncoding())
           } catch (_) {
             throw error
           }
--- a/src/text-buffer.js
+++ b/src/text-buffer.js
@@ -1939,14 +1939,14 @@ class TextBuffer {
             throw error
           }
         } else if (isWindows) {
-          const winattr = require('winattr')
-          const attrs = winattr.getSync(filePath)
+          const winattr = getPromisifiedWinattr()
+          const attrs = await winattr.get(filePath)
           if (!attrs.hidden) throw error
 
           try {
-            winattr.setSync(filePath, { hidden: false })
+            await winattr.set(filePath, { hidden: false })
             await this.buffer.save(filePath, this.getEncoding())
-            winattr.setSync(filePath, { hidden: true })
+            await winattr.set(filePath, { hidden: true })
           } catch (_) {
             throw error
           }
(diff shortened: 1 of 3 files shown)


src/text-buffer.js
```diff
@@ -1927,16 +1927,29 @@ class TextBuffer {
...
           try {
             await this.buffer.save(fsAdmin.createWriteStream(filePath), this.getEncoding())
           } catch (_) {
             throw error
           }
-        } else {
-          throw error
+        } else if (isWindows) {
+          const winattr = require('winattr')
+          const attrs = winattr.getSync(filePath)
```
aminya, line 1943: Shouldn't this be async so it does not block the main thread?
```suggestion
          const attrs = await winattr.get(filePath)
```

src/text-buffer.js
```diff
@@ -1927,16 +1927,29 @@ class TextBuffer {
...
           const fsAdmin = require('fs-admin')
           try {
             await this.buffer.save(fsAdmin.createWriteStream(filePath), this.getEncoding())
           } catch (_) {
             throw error
           }
-        } else {
-          throw error
+        } else if (isWindows) {
+          const winattr = require('winattr')
```
aminya, line 1942: You can cache `winattr` to speed up the execution. Just define `winattr` above the file:
```
let winattr = null
```

```suggestion
		  if (!winattr) {
		    winattr = require('winattr')
		  }
```

src/text-buffer.js
```diff
@@ -1927,16 +1927,29 @@ class TextBuffer {
...
+          const attrs = winattr.getSync(filePath)
+          if (!attrs.hidden) throw error
+
+          try {
+            winattr.setSync(filePath, { hidden: false })
```
aminya, line 1947:
```suggestion
            await winattr.set(filePath, { hidden: false })
```

src/text-buffer.js
```diff
@@ -1927,16 +1927,29 @@ class TextBuffer {
...
+            await this.buffer.save(filePath, this.getEncoding())
+            winattr.setSync(filePath, { hidden: true })
```
aminya, line 1949:
```suggestion
            await winattr.set(filePath, { hidden: true })
```

src/text-buffer.js
gosukiwi: `winattr.get` does not return a promise, it expects a callback, so I can't use `await` there. 

Also, I'm not sure it's using an async API at all, so it could still be blocking. I opted to use the `sync` variants to avoid the callback hell :P given saving a hidden file will rarely be a bottle neck I think it's a fair use.

src/text-buffer.js
aminya: @gosukiwi Check here. I have promisified the functions. I have submitted a PR to the original repository:
https://github.com/stevenvachon/winattr/pull/12

We can either wait for that, or we promisify the functions ourselves.

gosukiwi: For that I copied the way the original code does it:

```javascript
if (isMacOrLinux) {
  const fsAdmin = require('fs-admin')
  // ...
} else (isWindows) {
  const winattr = require('winattr')
}
```

I like the fact that it's lazy and wont be required unless needed, besides, node will cache it later.

aminya: We need to cache the promisified functions though.

gosukiwi: Does Atom provide promisify? I don't think introducing the `promisify` dependency is a good idea otherwise. Besides, what advantages are we really getting with it? I'd rather have simple code, and measure before trying to optimize.

gosukiwi: Ohh does `promisify` lives in `util`?

aminya:
Sorry for going back and forth. I have updated the original winattr. If we wait for that a little, we can use the promisfied library without adding extra code.

https://github.com/stevenvachon/winattr/pull/12

gosukiwi:
@aminya no worries. I think it might be work merging this anyways, though. Your change is a breaking change and I don't expect the author to merge it soon. But who knows :P I can always revert this commit if needed, so no big deal.

Let's see what the maintainers say.

src/text-buffer.js
```diff
@@ -1927,16 +1927,29 @@ class TextBuffer {
...
+        } else if (isWindows) {
+          const winattr = getPromisifiedWinattr()
+          const attrs = await winattr.get(filePath)
+          if (!attrs.hidden) throw error
+
+          try {
+            await winattr.set(filePath, { hidden: false })
+            await this.buffer.save(filePath, this.getEncoding())
+            await winattr.set(filePath, { hidden: true })
+          } catch (_) {
```
aminya, line 1950:
```suggestion
          } catch (error) {
```

gosukiwi: I don't think we want to do this, as we want to throw the original error if something goes wrong

src/text-buffer.js
```diff
@@ -2601,4 +2614,18 @@ class SearchCallbackArgument {
   }
 }
 
+let _winattr = null
```
sadick254, line 2617: Do we really need this here?

src/text-buffer.js
sadick254, line 1950: What do you think about only checking for `windows` and put the rest of the code in an `else` block
Something like 
```js
if(isWindows) {
  const winattr = getPromisifiedWinattr()
  const attrs = await winattr.get(filePath)
  if (!attrs.hidden) throw error

  try {
    await winattr.set(filePath, { hidden: false })
    await this.buffer.save(filePath, this.getEncoding())
    await winattr.set(filePath, { hidden: true })
  } catch (_) {
    throw error
  }
} else {
  const fsAdmin = require('fs-admin')
  try {
    await this.buffer.save(fsAdmin.createWriteStream(filePath), this.getEncoding())
  } catch (_) {
    throw error
  }
}
```

sadick254 (commented):
Looks great. Just 2 comments.

sadick254:
>  If we wait for that a little, we can use the promisfied library without adding extra code.

We could create a follow up PR once @aminya 's  [PR](https://github.com/stevenvachon/winattr/pull/12) is merged we don't have to wait since we don't know when the PR will be done.

src/text-buffer.js
gosukiwi: Makes total sense :) I'll add a fix later today

src/text-buffer.js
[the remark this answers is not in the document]
gosukiwi: Do you want it to live somewhere else in the file? Maybe at the top? I added it there because in order to understand the `getPromisifiedWinattr` function you can just look at that piece of code and understand it all, no need to scroll to know what `_winattr` is. But I'm happy to move it somewhere else to follow Atom's conventions.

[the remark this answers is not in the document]
sadick254: I wasn't getting the idea behind defining it there. But I think I see the point now. Correct me if am work. It is defined here so that calling `getPromisifiedWinattr` multiple times does not result in multiple `require` calls.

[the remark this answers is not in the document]
gosukiwi: Yeah that's right. It will also only promisify the methods once.

aminya:
I think you should go ahead and merge this. I will create a new promisified `winattr` packages that remove the callbacks. In https://github.com/stevenvachon/winattr/pull/12, @stevenvachon is not open to removing the callbacks, which is the only reason I made that.

gosukiwi:
Thanks for the review @sadick254! I addressed your comments.

sadick254 (approved):
Looks great. 👍

PR #336, opened 2020-12-16
Bump winattr to @aminya/winattr@4.0.3

aminya:
### Description of the change

This replaces winattr with the promisified version which is registered under my name. 
https://github.com/aminya/winattr

This also bumps fswin, which may solve the issue of fswin.node in Linux/MacOS for Electron upgrade.  https://github.com/atom/atom/pull/21777

### Benefits
- Promisifed 
- fswin bumped

### Verification 
Tests pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "serializable": "^1.0.3",
     "superstring": "^2.4.4",
     "underscore-plus": "^1.0.0",
-    "winattr": "^3.0.0"
+    "winattr": "npm:@aminya/winattr@^4.0.3"
   },
   "standard": {
     "env": {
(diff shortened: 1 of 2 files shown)


sadick254:
This doesn't resolve the issue on `linux` and `macOS` I still experience the same error while trying to build atom.
Screenshot 2020-12-17 at 08 52 18

aminya:
@sadick254 Should we just add copy-paste the error message and ignore it when we hit it? fswin binary is not used on Mac or Linux.

https://github.com/atom/atom/blob/f1b0e3b47b30ed9379821b341b3337696b8a2a12/script/lib/dump-symbols.js#L31

sadick254:
@aminya We can do that and add a comment. That would work.

PR #338, opened 2021-06-16
Migrate to Github Actions

sadick254:


sadick254:
The failure on macos is due to a flaky test that should be resolved soon.

